PR #6, opened 2018-01-24
ODATA-1099 - resolved - Add annotations to describe custom query options and custom headers

ralfhandl:
https://issues.oasis-open.org/browse/ODATA-1099


ralfhandl:
@mikepizzo @madansr7 please check:
- Aligned with https://github.com/madansr7/odata-vocabularies/blob/master/vocabularies/Org.OData.Capabilities.V1.xml
- Aligned with https://github.com/oasis-tcs/odata-vocabularies/pull/16 and https://issues.oasis-open.org/browse/ODATA-1200

Open:
- Do we need to differentiate RetrieveRestrictions between "get single" and "get list" for entity sets?

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -525,6 +525,12 @@ supported:
         <Property Name="QueryOptions" Type="Capabilities.ModificationQueryOptionsType" Nullable="true">
           <Annotation Term="Core.Description" String="Support for query options with insert requests" />
         </Property>
+        <Property Name="CustomHeaders" Type="Collection(Capabilities.CustomParameter)">
```
madansr7: can we use the name 'RequestHeaders' instead?

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -525,6 +525,12 @@ supported:
         <Property Name="QueryOptions" Type="Capabilities.ModificationQueryOptionsType" Nullable="true">
           <Annotation Term="Core.Description" String="Support for query options with insert requests" />
         </Property>
+        <Property Name="CustomHeaders" Type="Collection(Capabilities.CustomParameter)">
+          <Annotation Term="Core.Description" String="Supported or required custom headers" />
+        </Property>
+        <Property Name="CustomQueryOptions" Type="Collection(Capabilities.CustomParameter)">
```
madansr7: Suggesting to use name 'QueryParameters'.

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -661,6 +679,112 @@ supported:
...
+            String="This is only meaningful if the annotation is applied to an entity set or singleton." />
+        </Property>
+        <Property Name="RetrievableByKey" Type="Edm.Boolean" Nullable="false" DefaultValue="true">
+          <Annotation Term="Core.Description" String="Entities can be retrieved by key" />
+          <Annotation Term="Core.LongDescription" String="This is only meaningful if the annotation is applied to an entity set." />
+        </Property>
+        <Property Name="CustomHeaders" Type="Collection(Capabilities.CustomParameter)">
+          <Annotation Term="Core.Description" String="Supported or required custom headers" />
+        </Property>
+        <Property Name="CustomQueryOptions" Type="Collection(Capabilities.CustomParameter)">
```
madansr7: Suggesting to use name 'QueryParameters' for this property.

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -661,6 +679,112 @@ supported:
         </Property>
       </ComplexType>
 
+      <Term Name="RetrieveRestrictions" Type="Capabilities.RetrieveRestrictionsType" AppliesTo="EntitySet Singleton FunctionImport">
```
madansr7: I suggest 'ReadRestrictions' since the other restrictions are named 'update', 'insert' and 'delete' which match with the CRUD naming.

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -661,6 +679,112 @@ supported:
...
+        </Property>
+        <Property Name="CustomHeaders" Type="Collection(Capabilities.CustomParameter)">
+          <Annotation Term="Core.Description" String="Supported or required custom headers" />
+        </Property>
+        <Property Name="CustomQueryOptions" Type="Collection(Capabilities.CustomParameter)">
+          <Annotation Term="Core.Description" String="Supported or required custom query options" />
+        </Property>
+      </ComplexType>
+
+      <Term Name="CustomHeaders" Type="Collection(Capabilities.CustomParameter)" AppliesTo="EntityContainer">
```
madansr7: can we use the name 'RequestHeaders' instead?

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -661,6 +679,112 @@ supported:
...
+          <Annotation Term="Core.Description" String="Entities can be retrieved by key" />
+          <Annotation Term="Core.LongDescription" String="This is only meaningful if the annotation is applied to an entity set." />
+        </Property>
+        <Property Name="CustomHeaders" Type="Collection(Capabilities.CustomParameter)">
+          <Annotation Term="Core.Description" String="Supported or required custom headers" />
+        </Property>
+        <Property Name="CustomQueryOptions" Type="Collection(Capabilities.CustomParameter)">
+          <Annotation Term="Core.Description" String="Supported or required custom query options" />
+        </Property>
+      </ComplexType>
```
madansr7: The property 'QueryOptions' (type: ModificationQueryOptionsType) property could be added here to this term as well.

vocabularies/Org.OData.Capabilities.V1.xml
ralfhandl: How about RequestParameters to match the naming proposal for RequestHeaders (instead of CustomHeaders)?

vocabularies/Org.OData.Capabilities.V1.xml
ralfhandl: Sure. Hope that this does not mislead service designers to annotate OData-Version and OData-MaxVersion with this term.

vocabularies/Org.OData.Capabilities.V1.xml
ralfhandl: The R in CRUD is (at least at SAP) mostly interpreted as "read single" or "get by key", whereas "get list" is referred to as Q, leading to CRUDQ. Which also matches OpenAPI where "get by key" and "get list" are different path items with different sets of parameters (headers and query options), different descriptions, ....

"Retrieve" was chosen to avoid that possible confusion and to mean "R and Q".

vocabularies/Org.OData.Capabilities.V1.xml
ralfhandl: I don't think ModificationQueryOptions make sense for Read or RetrieveRestrictions. They would also clash with existing FilterRestrictions/Filterable, SortRestrictions/Sortable, etc.

vocabularies/Org.OData.Capabilities.V1.xml
ralfhandl: Changed to ReadRestrictions as decided in TC meeting on 2019-02-14, also changed property Retrievable to Readable.

Descriptions still use "retrieve" to match wording of Part 1: Protocol

PR #28, opened 2018-10-25
Fix ODATA-1247

mikepizzo:
add additional targets for DerivedTypeConstraint


mikepizzo:
Closing as dupe of ODATA-1232.

PR #30, opened 2018-11-16
ODATA-1064 - open - Add ability to annotate collections to return only count and NextLink

ralfhandl:


ralfhandl:
2019-03-21 close without action

PR #45, opened 2019-04-03
ODATA-1005 - open - Make sure we have capabilities for all new 4.01 functionality

ralfhandl:


vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -327,13 +328,20 @@ supported:
         <Annotation Term="Core.Description" String="Supports $skip" />
       </Term>
 
+      <Term Name="ComputeSupported" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="EntitySet">
```
mikepizzo: shouldn't all of these be nullable?

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -558,6 +571,18 @@ supported:
           String="Supports [key-as-segment convention](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_KeyasSegmentConvention) for addressing entities within a collection" />
       </Term>
 
+      <Term Name="RelaxedSystemQueryOptionNamesSupported" Type="Core.Tag" Nullable="false" DefaultValue="true"
```
mikepizzo: Do we need this? I would just say it's required for 4.01, but that clients should use canonical formats for interoperability. I can't imagine a client querying this and then adjusting behavior based on the value.

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -650,6 +679,12 @@ The absence of `RestrictedProperties` denotes all properties are accessible usin
         <Property Name="Updatable" Type="Edm.Boolean" Nullable="false" DefaultValue="true">
           <Annotation Term="Core.Description" String="Entities can be updated" />
         </Property>
+        <Property Name="Upsertable" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
+          <Annotation Term="Core.Description" String="Entities can be upserted" />
+        </Property>
+        <Property Name="CollectionUpdate" Type="Capabilities.CollectionUpdateType" Nullable="false">
+          <Annotation Term="Core.Description" String="Entities can be upserted" />
```
mikepizzo: Is this the right description?

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -650,6 +679,12 @@ The absence of `RestrictedProperties` denotes all properties are accessible usin
         <Property Name="Updatable" Type="Edm.Boolean" Nullable="false" DefaultValue="true">
           <Annotation Term="Core.Description" String="Entities can be updated" />
         </Property>
+        <Property Name="Upsertable" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
+          <Annotation Term="Core.Description" String="Entities can be upserted" />
+        </Property>
+        <Property Name="CollectionUpdate" Type="Capabilities.CollectionUpdateType" Nullable="false">
```
mikepizzo: >Nullable="false" [](start = 83, length = 16)

Should not be nullable=false

ralfhandl: Should be Nullable=true - check other complex properties as well

vocabularies/Org.OData.Capabilities.V1.xml
ralfhandl: No, we don't need it

vocabularies/Org.OData.Capabilities.V1.xml
ralfhandl: Should be nullable

vocabularies/Org.OData.Capabilities.V1.xml
ralfhandl: No, is PATCH on collection using delta format - with differentiation "only upsert" and "upsert and delete"

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -678,6 +713,24 @@ The absence of `RestrictedProperties` denotes all properties are accessible usin
           <Annotation Term="Core.IsLanguageDependent" />
         </Property>
       </ComplexType>
+      <ComplexType Name="CollectionUpdateType">
```
ralfhandl: Flatten this, move these four properties to UpdateRestrictionsType, mention in Description that they only are valid if applied to a collection

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -678,6 +713,24 @@ The absence of `RestrictedProperties` denotes all properties are accessible usin
           <Annotation Term="Core.IsLanguageDependent" />
         </Property>
       </ComplexType>
+      <ComplexType Name="CollectionUpdateType">
+        <Property Name="DeltaUpsertable" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
```
ralfhandl: Just call this DeltaUpdateSupported

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -678,6 +713,24 @@ The absence of `RestrictedProperties` denotes all properties are accessible usin
           <Annotation Term="Core.IsLanguageDependent" />
         </Property>
       </ComplexType>
+      <ComplexType Name="CollectionUpdateType">
+        <Property Name="DeltaUpsertable" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
+          <Annotation Term="Core.Description"
+            String="Entities can be inserted and updated via a PATCH request with a delta payload" />
+        </Property>
+        <Property Name="DeltaDeletable" Type="Edm.Boolean" Nullable="false" DefaultValue="true">
```
ralfhandl: Remove this

vocabularies/Org.OData.Capabilities.V1.xml
ralfhandl: Removed and moved properties of CollectionUpdateType here

Issue #53, opened 2019-06-19
Permission should be collection of (PermissionType) in Capabilities


xuzhg:
See: https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Capabilities.V1.xml#L580

it should be:
```xml
<Property Name="Permission" Type="Collection(Capabilities.PermissionType)" Nullable="true">
```

And it's better to rename as "Permissions".

xuzhg:
@madansr7 @ralfhandl @mikepizzo

PR #54, opened 2019-06-19
ODATA-1316 - applied - Should support multiple permissions for read/write

madansr7:
updating permissions sample, property name and type to address the issue logged via issue [https://github.com/oasis-tcs/odata-vocabularies/issues/53](https://github.com/oasis-tcs/odata-vocabularies/issues/53)

### DIFF
--- a/examples/Org.OData.Capabilities.V1.permissions-sample.xml
+++ b/examples/Org.OData.Capabilities.V1.permissions-sample.xml
@@ -9,102 +9,106 @@
         <Annotation Term="Capabilities.InsertRestrictions">
           <Record>
             <PropertyValue Property="Permissions">
-              <Record>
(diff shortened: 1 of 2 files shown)


xuzhg:
Thanks for updating.  I have another misunderstanding

https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Capabilities.V1.xml#L625

`<Property Name="Scheme" Type="Auth.SecurityScheme" Nullable="false">`

should be:

`<Property Name="Scheme" Type="Edm.String" Nullable="false">`

If it's correct, please update it also.

madansr7:
@mikepizzo , let me know if this warrants a separate issue and PR.

PR #55, opened 2019-06-21
ODATA-1299 - resolved - Continue-on-Error for Collection requests

ralfhandl:


vocabularies/Org.OData.Core.V1.json
```diff
@@ -235,6 +235,97 @@
...
+                "$Nullable": true,
+                "$Type": "Core.Message",
+                "@Core.Description": "Information on the (cause of) the exception"
+            }
+        },
+        "ValueExceptionType": {
+            "$Kind": "ComplexType",
+            "$BaseType": "Core.ExceptionType",
+            "value": {
+                "@Core.Description": "The actual numeric value"
```
Hubert-Heijkers: The value property is optional right, so isn't there a "$Nullable": true missing here?

vocabularies/Org.OData.Core.V1.json
```diff
@@ -235,6 +235,97 @@
...
+            "$Kind": "Term",
+            "$Nullable": true,
+            "$Type": "Core.NestedItemExceptionType",
+            "@Core.Description": "The annotated instance or collection within a success payload is problematic"
+        },
+        "NestedItemExceptionType": {
+            "$Kind": "ComplexType",
+            "$BaseType": "Core.ExceptionType",
+            "retryLink": {
+                "@Core.Description": "GET this URL to retry retrieving the problematic instances",
```
Hubert-Heijkers: Same here, retryLink is optional, $Nullable: true is missing

vocabularies/Org.OData.Core.V1.json
```diff
@@ -235,6 +235,97 @@
                 }
             ]
         },
+        "ValueException": {
+            "$Kind": "Term",
+            "$Nullable": true,
```
Hubert-Heijkers: Can a term be nullable?

PR #56, opened 2019-06-26
ODATA-1318 - resolved - Fix issues with Capabilities vocabulary

madansr7:
- Updated the capabilities.xml to change property `Scheme' under complex type `PermissionType` to be of type 'Edm.String'.

  This was causing us to render the entire Auth.Scheme node as a nested element inside the Read/Insert/Update/Delete Restrictions. The change to Edm.String will allow naming the AuthScheme that is available and targeted at the container. Auth schemes defined at the container level contain properties such as `authurl`, `refreshurl`, etc. from the auth vocabulary.

- Updated `OperationRestrictions` to remove 'Collection'. It should be only a single record of `Capabilities.OperationRestrictionsType`.

  OperationRestrictions does not need to be a collection type since we have no way to identify the parameters (in case of overload) for which this would be a collection. There is already a mechanism that exists to create annotations to target specific overloads, so collection is needless.

- Renamed `OperationRestrictions` to `OperationRestrictionsType` to be consistent with naming conventions in the vocabulary.

- Removed `FunctionImport` from ReadRestrictions.  

  FunctionImport has been moved to OperationRestrictions because the ReadByKeyRestriction on the FunctionImport should be put on the corresponding entityset. 
To restrict the FunctionImport we should use the OperationRestrictions term.

- Added `FunctionImport` and `ActionImport` to OperationRestrictions.

  Added FunctionImport and ActionImport based on the removal of FunctionImport from ReadRestrictions.

### DIFF
--- a/vocabularies/Org.OData.Authorization.V1.xml
+++ b/vocabularies/Org.OData.Authorization.V1.xml
@@ -72,7 +72,7 @@
       </Term>
 
       <ComplexType Name="SecurityScheme">
-        <Property Name="Authorization" Type="Edm.String" Nullable="false">
+        <Property Name="Authorization" Type="Auth.SchemeName" Nullable="false">
           <Annotation Term="Core.Description" String="The name of a required authorization scheme" />
         </Property>
         <Property Name="RequiredScopes" Type="Collection(Edm.String)" Nullable="false">
@@ -187,6 +187,10 @@
           <Annotation Term="Core.Description" String="API Key is passed as a cookie" />
         </Member>
       </EnumType>
+
+      <TypeDefinition Name="SchemeName" UnderlyingType="Edm.String">
+          <Annotation Term="Core.Description" String="The name of the auth scheme." />
+      </TypeDefinition>
 
     </Schema>
   </edmx:DataServices>
--- a/vocabularies/Org.OData.Authorization.V1.xml
+++ b/vocabularies/Org.OData.Authorization.V1.xml
@@ -189,7 +189,7 @@
       </EnumType>
 
       <TypeDefinition Name="SchemeName" UnderlyingType="Edm.String">
-          <Annotation Term="Core.Description" String="The name of the auth scheme." />
+          <Annotation Term="Core.Description" String="The name of the authorization scheme." />
       </TypeDefinition>
 
     </Schema>
--- a/vocabularies/Org.OData.Capabilities.V1.xml
+++ b/vocabularies/Org.OData.Capabilities.V1.xml
@@ -600,7 +600,7 @@ supported:
             String="Entities of a specific derived type can be created by specifying a type-cast segment" />
         </Property>
         <Property Name="Permissions" Type="Collection(Capabilities.PermissionType)" Nullable="true">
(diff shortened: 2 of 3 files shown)


ralfhandl:
Does this pull request replace https://github.com/oasis-tcs/odata-vocabularies/pull/54, or does it depend on it?

In either case: why not combine them into one PR?

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -622,7 +622,7 @@ supported:
       </ComplexType>
 
       <ComplexType Name="PermissionType">
-        <Property Name="Scheme" Type="Auth.SecurityScheme" Nullable="false">
+        <Property Name="Scheme" Type="Edm.String" Nullable="false">
```
ralfhandl: If this corresponds to 

Authorization | String | The name of a required authorization scheme
-- | -- | --

Then it should better be named `AuthorizationSchemeName` or just `SchemeName` because it is defined in the Authorization vocabulary..

@mikepizzo instead of relying on names we could add a type definition `SchemeName` of type `String` to the Authorization vocabulary and use this type definition consistently here and within the Authorization vocabulary.

What do you think?

mikepizzo: I had the same idea of defining a SchemeName type definition to be more self-documenting, although it does add one more type to resolve. We can discuss in our F2F.

madansr7:
1) Created `SchemeName` typedefinition in auth vocabulary
2) Referred Auth.SchemeName in Auth vocabulary and Capability vocabulary.
3) Changed Permission property to be Collection(Capabilities.PermissionType)
4) In PermissionType, changed Property name `Scheme` to `SchemeName`
5) For Property `Permissions`, updated the Type to `Auth.SchemeName`
4) Changed Property name `Permission` to `Permissions` to match naming convention since its a collection.
5) Removed `ActionImport` and `FunctionImport` from OperationRestriction
6) Removed `FunctionImport` from ReadRestrictions
7) Removed Collection from OperationRectriction.

vocabularies/Org.OData.Authorization.V1.xml
```diff
@@ -188,6 +188,10 @@
         </Member>
       </EnumType>
 
+      <TypeDefinition Name="SchemeName" UnderlyingType="Edm.String">
+          <Annotation Term="Core.Description" String="The name of the auth scheme." />
```
ralfhandl: Minor: please do not abbreviate, use "authorization scheme"

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -622,7 +622,7 @@ supported:
       </ComplexType>
 
       <ComplexType Name="PermissionType">
-        <Property Name="Scheme" Type="Auth.SecurityScheme" Nullable="false">
+        <Property Name="SchemeName" Type="Auth.SchemeName" Nullable="false">
           <Annotation Term="Core.Description" String="Auth flow scheme name" />
```
ralfhandl: "Authorization flow scheme name", no abbrev :-)

PR #73, opened 2020-04-24
ODATA-1365 - open - Symbolic name for allowed values

ralfhandl:


ralfhandl:
Fixed, type is now `Edm.String`

ralfhandl:
Changed

PR #74, opened 2020-05-15
ODATA-839 - resolved -A recursive hierarchy annotation may also contain a navigation property ...

GeraldKrause:
A recursive hierarchy annotation may also contain a navigation property to the children nodes


GeraldKrause:
This PR was replaced by https://github.com/oasis-tcs/odata-vocabularies/pull/121, which contains the fixes for [ODATA-1390](https://issues.oasis-open.org/browse/ODATA-1390)

PR #75, opened 2020-05-15
[ODATA-1367] split properties markdown table into one for properties and one for terms

chrisspre:
- split the single call to propertyTable in function typeSection into to
- moved functionality from propertyTable and propertyLines pertaining applicable terms into new functions applicableTermsTable and applicableTermLines respectively.
- changed layout of applicableTermsTable to two columns
- re-generated capabilities and Validation markdown files

### DIFF
--- a/lib/csdl2markdown.js
+++ b/lib/csdl2markdown.js
@@ -351,10 +351,10 @@ module.exports.csdl2markdown = function (filename, csdl) {
         const tLines = applicableTermLines(type);
         if (tLines.length > 0) {
             lines.push('');
-            lines.push('Notable terms');
+            lines.push('**Notable terms:**');
             lines.push('');
-            lines.push('Term|Type');
-            lines.push(':-------|:---');
+            // lines.push('Term|Type');
+            // lines.push(':-------|:---');
             lines.push(...tLines);
         }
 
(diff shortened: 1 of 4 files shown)


lib/csdl2markdown.js
```diff
@@ -335,14 +336,54 @@ module.exports.csdl2markdown = function (filename, csdl) {
...
+     */
+    function applicableTermsTable(type) {
+        const lines = [];
+
+        const tLines = applicableTermLines(type);
+        if (tLines.length > 0) {
+            lines.push('');
+            lines.push('Notable terms');
+            lines.push('');
+            lines.push('Term|Type');
```
ralfhandl, line 356:
```suggestion
            lines.push('Term|Link to Definition');
```

Or we could make the `@Core.Description` etc. the link and not repeat the substring. Then maybe use a simple bullet list instead of a table with only one column.

chrisspre: Updated to a markdown format that looks like:
```
- [Description](Org.OData.Core.V1.md#Description)
- [LongDescription](Org.OData.Core.V1.md#LongDescription)
```

lib/csdl2markdown.js
```diff
@@ -335,14 +336,53 @@ module.exports.csdl2markdown = function (filename, csdl) {
...
+     * @param {object} type Structured type
+     * @return {Array} Array of strings containing Markdown lines
+     */
+    function applicableTermsTable(type) {
+        const lines = [];
+
+        const tLines = applicableTermLines(type);
+        if (tLines.length > 0) {
+            lines.push('');
+            lines.push('**Notable terms:**');
```
ralfhandl, line 354:
```suggestion
            lines.push('**Applicable Terms:**');
```

One last change: as this list is the contents of term "ApplicableTerms", why not use that as the headline.

ralfhandl:
@chrisspre one more thing: please bump the version in package.json to 0.3.0

Rhanks!

PR #83, opened 2020-07-02
Contacts change ...

sthagen:
@chet-ensign ... this change might be applicable to many Projects under the oasis-tcs ☂ organization


OASIS-OP-Admin:
Thanks for bringing that to my attention Stefan. I agree that I should check on other TC GitHubs. I'm sure many are this way.

PR #108, opened 2021-03-10
Fix: prettiered and linted

ralfhandl:


HeikoTheissen (approved):
I verified that the actual vocabularies are not affected by this change.

PR #110, opened 2021-03-17
Show descriptions for parameters and return types in markdown

HeikoTheissen:


ralfhandl (changes requested):
I like the idea in principle, but the current output is unreadable.

Maybe we should have a subsection for each function similar to what we do for structured types, with the function description after the section header as prose text, and a table for the parameters with columns "Name", "Type", and "Description". The return type could be the last line of that table with `&rarr;` in the "Name" column.

ralfhandl:
Proposal:

## Functions

### [isroot](Org.OData.Aggregation.V1.xml#L181)

Returns true, if and only if the value of the node property of the specified hierarchy is the root of the hierarchy

Parameter|Type|Description
:-------|:--------|:----------
Entity | EntityType | ...
Hierarchy | String | ...
&rarr; | Boolean | ...

### [isdescendant](Org.OData.Aggregation.V1.xml#L189)

Returns true, if and only if the value of the node property of the specified hierarchy is a descendant of the given parent node with a distance of less than or equal to the optionally specified maximum distance

...

ralfhandl (changes requested):
Still unreadable, please see my proposal: 
- One section on level 3 for each function
- Description of function as paragraph
- Table of parameters and return type

ralfhandl (approved):
Nice!

Now the only piece of information missing is the " bound" indicator.

That could be a paragraph below the headline saying

### UpdateObject

**Bound:** true

Or the name of the first parameter could be in _italic_ - maybe a bit too subtle.

Or the description of the binding parameter could be prefixed:

Parameter|Type|Description
:------- |:---|:----------
|timeslices | [Entity] | **Binding Parameter**: Time slices to modify

ralfhandl (approved):

PR #114, opened 2021-03-19
Extension of markdown vocabularies

HeikoTheissen:
- [x] Discuss in TC on 2021-03-25
- [ ] Discuss in TC on 2021-04-15
* Source links for parameters and return type
* Deprecated parameters
* ~~Mark a term that "AppliesViaContainer" or "AppliesToTypeIfDynamic"~~ (rejected by TC since we don't show AppliesTo)
* Show "AllowedTerms" for AnnotationPaths

TC suggestions:
* name of binding parameter in bold, definition starts with "Binding parameter" (in bold)
* name of optional parameter in italics, definition starts with "Optional parameter" (in italics)


ralfhandl (approved):

package.json
```diff
@@ -17,7 +17,7 @@
   "main": "lib/csdl2markdown.js",
   "dependencies": {
     "minimist": "^1.2.0",
-    "odata-csdl": "^0.2.4"
+    "odata-csdl": "github:oasis-tcs/odata-csdl-schemas"
```
HeikoTheissen, line 20: I changed this only temporarily so that I can push before we have updated npm. Will change back to `"odata-csdl": "^0.2.6"` before merging.

ralfhandl: 0.2.6 is available on https://www.npmjs.com/package/odata-csdl

chrisspre (approved):

PR #118, opened 2021-03-25
Temporal: adapted to latest working draft

ralfhandl:
- Remove system time
- Support time-slice collections per temporal object


vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -69,22 +69,19 @@
 
       <!-- Metadata annotations -->
 
-      <Term Name="TemporalSupport" Type="Temporal.TemporalSupportType" AppliesTo="EntitySet NavigationProperty">
+      <Term Name="TemporalSupport" Type="Temporal.TemporalSupportType" Nullable="false" AppliesTo="EntitySet NavigationProperty">
```
HeikoTheissen, line 72:
```suggestion
      <Term Name="TemporalSupport" Type="Temporal.TemporalSupportType" Nullable="false" AppliesTo="Collection">
        <Annotation Term="Core.AppliesViaContainer"/>
```

examples/Org.OData.Temporal.V1-history-sample.xml
```diff
@@ -0,0 +1,101 @@
...
+        <EntitySet Name="Employees" EntityType="OrgModel.Employee" />
+        <EntitySet Name="Departments" EntityType="OrgModel.Department" />
+      </EntityContainer>
+
+      <Annotations Target="OrgModel.Default/Employees/history">
+        <Annotation Term="Temporal.TemporalSupport">
+          <Record>
+            <PropertyValue Property="ApplicationTime">
+              <Record Type="Temporal.GranularityDate">
+                <PropertyValue Property="SupportedQueries" EnumMember="Temporal.QueryType/TimeTravel Temporal.QueryType/TimeSeries" />
```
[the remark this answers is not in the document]
HeikoTheissen: But in both cases the time travel request URL `Employees?$at=2021-05-01` is allowed and that never mentions the `history` navigation property. Shouldn't the time travel annotation then also be targeted at the `Employees` entity set in both cases?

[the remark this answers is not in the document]
ralfhandl: In the "history" example the results of accessing the root entity sets will not depend on the value of `$at`, only the expanded or directly accessed `history` navigation will be affected.

So the precise requests are

~~~http
GET /Employees?$expand=history($at=2021-05-01)
GET /Employees/123/history?$at=2021-05-01
~~~

The "sloppy" request
~~~http
GET /Employees?$at=2021-05-01&$expand=history
~~~
relies on `$at` being cascaded along navigation.

In the unexpanded request
~~~http
GET /Employees?$at=2021-05-01
~~~
the `$at` has no effect.

HeikoTheissen (changes requested):
See my comment above.

examples/Org.OData.Temporal.V1-history-sample.xml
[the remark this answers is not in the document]
HeikoTheissen: Sorry, I meant `Employees('Heiko')?$at=2021-05-01`. This is allowed in both examples, right?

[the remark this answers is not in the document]
ralfhandl: > Shouldn't the time travel annotation then also be targeted at the `Employees` entity set in both cases?

No: time travel has no effect on `Employees`, it only has an effect on `Employees/history`.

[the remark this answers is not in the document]
ralfhandl: > This is allowed in both examples, right?

This is allowed in both examples, but it has no effect in the "history" example. We annotate where a client can see an effect.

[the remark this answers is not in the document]
HeikoTheissen: So if I want to see my data as of May 1st:
* History example: `Employees('Heiko')/history?$at=2015-05-01` returns properties From, To, Name and JobTitle
* Snapshot example: `Employees('Heiko')?$at=2015-05-01` returns properties ID, Name and JobTitle

Correct?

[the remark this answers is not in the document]
ralfhandl: Correct.

For a "mixed-style" API where the root entity type also has "snapshot" properties of actually time-dependent values (as accidentally in the the "history" examples until a few minutes ago), then I'd expect the root entity set to also be annotated with `TemporalSupport`.

HeikoTheissen (approved):

examples/Org.OData.Temporal.V1-snapshot-sample.json
```diff
@@ -0,0 +1,77 @@
...
+            "Name": {},
+            "Employees": {
+                "$Kind": "NavigationProperty",
+                "$Collection": true,
+                "$Type": "OrgModel.Employee"
+            }
+        },
+        "Default": {
+            "$Kind": "EntityContainer",
+            "Employees": {
```
ralfhandl, line 54: Add navigation property bindings, also for Departments

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -111,54 +117,24 @@
         <Annotation Term="Core.Description" String="Granularity of this time dimension is Date" />
         <Property Name="ClosedClosedPeriods" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
           <Annotation Term="Core.Description" String="Application-time periods are stored with closed-closed semantics" />
-          <Annotation Term="Core.LongDescription" String="Use the temporal query option `$toInclusive` instead of `$to` for specifying the period end in time-series queries." />
+          <Annotation Term="Core.LongDescription" String="If `true`, the value of the property referenced by `TimeSliceEnd` is included in the time slice, and the temporal query option `$toInclusive` must be used instead of `$to` for specifying the period end in time-series queries." />
         </Property>
```
ralfhandl, line 121: Name this `EndInclusive`?

examples/Org.OData.Temporal.V1-history-sample.json
```diff
@@ -0,0 +1,133 @@
...
+                "$Collection": true,
+                "$Type": "OrgModel.Department"
+            }
+        },
+        "$Annotations": {
+            "OrgModel.Default/Employees/history": {
+                "@Temporal.TemporalSupport": {
+                    "ApplicationTime": {
+                        "@odata.type": "https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Temporal.V1.xml#Temporal.GranularityDate",
+                        "SupportedQueries": "PointInTime,TimePeriod",
```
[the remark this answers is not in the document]
ralfhandl: `SupportedActions` was accidentally nested within `ApplicationTime`, is actually next to it.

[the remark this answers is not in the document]
ralfhandl: Maybe we should replace this with the kind of the entity set:
- snapshot --> only point-in-time possible
- timeslice --> both types of queries possible (would be artificial to not support both), and Start, End are required
Could be done via sub-typing, but that would "multiply" with the granularity Date/Timestamp into four sub-types

HeikoTheissen (approved):

examples/Org.OData.Temporal.V1-history-sample.json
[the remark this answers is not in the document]
HeikoTheissen: You mean replace SupportedQueries with an enumerated property (Snapshot, Timeslice)? I support that.

vocabularies/Org.OData.Temporal.V1.xml
HeikoTheissen: Even if ClosedClosedPeriods = true in date granularity, can't I use `$from=2021-04-01&$to=2021-05-01` to cover the month of April? The choice between `$to` and `$toInclusive` should be independent of the ClosedClosedPeriods flag. Or: If it is not independent, we don't need `$toInclusive`, we need only `$to`, and whether it is inclusive depends on ClosedClosedPeriods.

ClosedClosedPeriods should only affect how `$at` is handled.

examples/Org.OData.Temporal.V1-history-sample.json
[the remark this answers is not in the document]
ralfhandl: SupportedQueries currently has an enumeration type:

https://github.com/oasis-tcs/odata-vocabularies/blob/80af54c275678f342e7316e84c9b2e2217fd6d08/vocabularies/Org.OData.Temporal.V1.xml#L91

https://github.com/oasis-tcs/odata-vocabularies/blob/80af54c275678f342e7316e84c9b2e2217fd6d08/vocabularies/Org.OData.Temporal.V1.xml#L102-L105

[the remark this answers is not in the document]
HeikoTheissen: But snapshot and timeslice should be mutually exclusive (not IsFlags="true").

vocabularies/Org.OData.Temporal.V1.xml
ralfhandl: You can use `$to` and `$toInclusive` independent of the server-side storage with closed-open or closed-closed periods, the server has to take care of how to translate these into queries to its underlying data store.

The annotation just helps explain the results to the client, and to correctly interpret the values of the `TimeSliceEnd` property.

examples/Org.OData.Temporal.V1-history-sample.json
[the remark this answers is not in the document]
ralfhandl: Entity sets are either "snapshot" or "timeslice".

But "timeslice" entity sets can support both "point-in-time" and "time-period" queries.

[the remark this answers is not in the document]
HeikoTheissen: But I would prefer an annotation on the entity set that makes "snapshot" vs. "timeslice" explicit. At the moment "Temporal.QueryType/TimePeriod" implies "timeslice", but I find this too implicit.

vocabularies/Org.OData.Temporal.V1.xml
HeikoTheissen: > You can use `$to` and `$toInclusive` independent of the server-side storage with closed-open or closed-closed periods

But the annotation definition says otherwise: If `ClosedClosedPeriod=true` then I must use `$toInclusive`.

examples/Org.OData.Temporal.V1-history-sample.json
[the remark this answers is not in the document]
ralfhandl: Good idea: replace `SupportedQueries` with `TemporalCollectionKind` and two (non-flag) values `Snapshot` and `Timeslice`.

`Kind=TimeSlice` then requires specifying `TimeSliceStart` and `TimeSliceEnd`.

vocabularies/Org.OData.Temporal.V1.xml
ralfhandl: Thanks, fixed

examples/Org.OData.Temporal.V1-history-sample.json
```diff
@@ -0,0 +1,133 @@
...
+            },
+            "Departments": {
+                "$Collection": true,
+                "$Type": "OrgModel.Department"
+            }
+        },
+        "$Annotations": {
+            "OrgModel.Default/Employees/history": {
+                "@Temporal.TemporalSupport": {
+                    "ApplicationTime": {
```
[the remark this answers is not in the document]
HeikoTheissen: What about `Temporal.ApplicationTimeSupport`?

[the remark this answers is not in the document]
ralfhandl: Done.

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,98 +67,82 @@
...
+
+      <ComplexType Name="CollectionKindTimeSlice">
+        <Annotation Term="Core.Description" String="Each OData entity corresponds to a time slice and all time slices of a temporal object are part of the collection" />
+        <Property Name="TimeSliceStart" Type="Edm.PropertyPath" Nullable="false">
+          <Annotation Term="Core.Description" String="Property containing lower boundary of a time slice" />
+          <Annotation Term="Core.LongDescription" String="Must be present if time-period queries are supported" />
+        </Property>
+        <Property Name="TimeSliceEnd" Type="Edm.PropertyPath" Nullable="false">
+          <Annotation Term="Core.Description" String="Property containing upper boundary of a time slice" />
+          <Annotation Term="Core.LongDescription" String="Must be present if time-period queries are supported" />
```
HeikoTheissen, line 109: You write "must be present if ...", but it is not nullable. Contradiction?

ralfhandl: Leftover, now guaranteed by inheritance

HeikoTheissen (approved):

HeikoTheissen:
What's the extra value in `$from`, `$to` and `$toInclusive`? If `Temporal.CollectionKindTimeSlice/TimeSliceStart` and `End` were `Edm.Path`s instead of `Edm.PropertyPath`s, the same could be expressed equivalently as
```
$filter=@Temporal.ApplicationTimeSupported/CollectionKind/TimeSliceStart<toDate
and @Temporal.ApplicationTimeSupported/CollectionKind/TimeSliceEnd>fromDate
```
without the additional syntax. And this variant is much more powerful, because you can freely choose between inclusive/exclusive boundaries, intersection/containment etc.

ralfhandl:
Implementation of new query options is easier than support for annotation expressions in $filter. Easiest implementation of $at etc. is to only allow literal values as "expressions".

Also `$from=2021-04-01&$to=2021-05-01` is easier than
* looking up in $metadata whether the collection is using closed-open or closed-closed semantics
* getting all comparison operators right

HeikoTheissen:
> This again needs adaptation of the $filter parser

But the $filter parser must already be able to parse expressions involving functions. And `Temporal.fromto` would be just a function.

ralfhandl:
Using a filter function makes it harder to propagate the temporal filter along $expand: you would have to repeat it on every expand clause for collection-valued navigation properties.

HeikoTheissen:
> Using a filter function makes it harder to propagate the temporal filter along $expand

I know that $at is propagated along $expand, this is described in section 4.2.1. But are $from and $to(Inclusive) also propagated along $expand? Nothing like this is not mentioned in section 4.2.2. But if it were, that would of course justify the extra syntax.

ralfhandl:
> Temporal query options can be specified even if the directly addressed entity set does not track time. In this case the temporal query options do not have any effect other than being “cascaded” along $expand. 

Means all four of the temporal query options, not restricted to $at.

> If the time-range query options $from and $to/$toInclusive are specified or cascaded along $expand, all other system query options on the same expansion level are evaluated on the time slices whose period overlaps with the requested period.

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,98 +67,79 @@
...
 
+      <!-- TODO: move to separate TemporalActions vocabulary? -->
       <!-- Operations on Temporal Objects -->
 
-      <Action Name="UpdateObject" IsBound="true" EntitySetPath="timeslices">
+      <Action Name="Update" IsBound="true" EntitySetPath="timeslices">
         <Annotation Term="Core.Description" String="Updates all time slices of a temporal object in a given period of application time after splitting partially overlapping time slices" />
         <Parameter Name="timeslices" Type="Collection(Edm.EntityType)" Nullable="false">
           <Annotation Term="Core.Description" String="Time slices to modify" />
         </Parameter>
```
HeikoTheissen, line 142: If `Employee_history` had a named entity set, then the action call `Employee_history_Set/Temporal.Update` would have the binding parameter `timeslices` contain more than one temporal object, but only one of them is updated. This should be made clearer in the text.

HeikoTheissen (changes requested):

vocabularies/Org.OData.Temporal.V1.xml
ralfhandl: Added to specification draft

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,98 +67,79 @@
...
 
 
-      <!-- Metadata annotations -->
-
-      <Term Name="TemporalSupport" Type="Temporal.TemporalSupportType" AppliesTo="EntitySet NavigationProperty">
+      <Term Name="ApplicationTimeSupport" Type="Temporal.ApplicationTimeSupportType" Nullable="false" AppliesTo="Collection">
+        <Annotation Term="Core.AppliesViaContainer" />
         <Annotation Term="Core.Description">
-          <String>This entity set or navigation property supports  temporal requests</String>
+          <String>This entity set or navigation property supports temporal requests</String>
```
HeikoTheissen, line 73:
```suggestion
          <String>This collection supports temporal requests</String>
```

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,98 +67,79 @@
...
+      <ComplexType Name="CollectionKindType" Abstract="true">
+        <Annotation Term="Core.Description" String="Kind of temporal collection" />
       </ComplexType>
 
-      <TypeDefinition Name="QualifiedActionName" UnderlyingType="Edm.String">
-        <Annotation Term="Core.Description" String="The qualified name of an action in scope." />
-        <Annotation Term="Core.LongDescription" String="*TODO: move this to Core, where we already have QualifiedTypeName and QualifiedTermName?*" />
-      </TypeDefinition>
+      <ComplexType Name="CollectionKindSnapshot" BaseType="Temporal.CollectionKindType">
+        <Annotation Term="Core.Description" String="Each OData entity corresponds to a temporal object and by default only shows data of the current time slice" />
```
HeikoTheissen, line 98:
```suggestion
        <Annotation Term="Core.Description" String="Each OData entity corresponds to a temporal object and by default only shows data of the current time slice" />
        <Annotation Term="Core.LongDescription" String="Each OData entity is uniquely identified by its [`ApplicationTimeSupport/ObjectKey`](#ApplicationTimeSupportType)" />
```
I want to make the relation between object key and (logical) entity key more explicit. A similar statement for time-slice collections is introduced with #128.

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,98 +67,79 @@
...
-          <Annotation Term="Core.LongDescription" String="A temporal object is a set of facts whose change over application and/or system time is tracked by the service. The entities of the annotated entity set are time slices of temporal objects. A temporal object is uniquely identified within this set by the values of the specified object key properties. Object key properties follow the same rules as entity key properties. If no object key is specified, the object key is identical to the entity key." />
+          <Annotation Term="Core.LongDescription" String="A temporal object is a set of facts whose change over application and/or system time is tracked by the service. The entities of the annotated collection are time slices of potentially multiple temporal objects. A temporal object is uniquely identified within this set by the values of the specified object key properties. Object key properties follow the same rules as entity key properties. If no object key is specified, the annotated collection only contains time slices for a single one temporal object." />
```
HeikoTheissen, line 86:
```suggestion
          <Annotation Term="Core.LongDescription" String="A temporal object is a set of facts whose changes over application time are tracked by the service. The entities in the annotated collection belong to potentially multiple temporal objects, and each temporal object is uniquely identified by the values of the specified object key properties. Object key properties follow the same rules as entity key properties. If no object key is specified, only a single temporal object belongs to the annotated collection." />
```
Avoid the term "time slice" here, because it is defined only for time-slice collections.

Remove "within this set", because this sounds like the set of primitive properties.

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,98 +67,79 @@
...
+          <String>This entity set or navigation property supports temporal requests</String>
         </Annotation>
       </Term>
 
-      <ComplexType Name="TemporalSupportType">
-        <Property Name="ApplicationTime" Type="Temporal.TimeDimensionType">
-          <Annotation Term="Core.Description" String="Application time granularity" />
+      <ComplexType Name="ApplicationTimeSupportType">
+        <Property Name="CollectionKind" Type="Temporal.CollectionKindType" Nullable="false">
+          <Annotation Term="Core.Description" String="Kind of temporal collection" />
```
HeikoTheissen, line 79: Collection kind = Kind of collection. Redundant?

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,98 +67,79 @@
...
         </Property>
-        <Property Name="SystemTime" Type="Temporal.GranularityDateTimeOffset">
-          <Annotation Term="Core.Description" String="System time granularity" />
+        <Property Name="PeriodKind" Type="Temporal.PeriodKindType" Nullable="false">
+          <Annotation Term="Core.Description" String="Kind of application time period" />
```
HeikoTheissen, line 82:
```suggestion
          <Annotation Term="Core.Description" String="How the boundaries of an application time period may be defined" />
```

HeikoTheissen (changes requested):

vocabularies/Org.OData.Temporal.V1.xml
ralfhandl: Moved property `ObjectKey` to type `CollectionKindTimeSlice` because it is only relevant there.

vocabularies/Org.OData.Temporal.V1.xml
ralfhandl: This is the description of the abstract base type. Should this description list/mention all sub-types?

vocabularies/Org.OData.Temporal.V1.xml
ralfhandl: It's not only the data type of the boundaries, it can also be additional information such as `ClosedClosedPeriods`

vocabularies/Org.OData.Temporal.V1.xml
HeikoTheissen: Let's consider more expressive names:
* CollectionKind -> DataHistory
* PeriodKind -> UnitOfTime

HeikoTheissen (approved):

vocabularies/Org.OData.Temporal.V1.xml
ralfhandl: Let's try that:
- DataHistory
  - subtype DataHistoryHidden (for snapshot entity sets)
    - no properties, just a "marker type" 
  - subtype DataHistoryVisible (for time-slice entity sets)
    - properties:
      - TimeSliceStart (rename to PeriodStart?)
      - TimeSliceEnd (rename to PeriodEnd?)
      - ObjectKey
  
- UnitOfTime
  - subtype UnitOfTimeDate
    - properties:
      - ClosedClosedPeriods
  - subtype UnitOfTimeDateTimeOffset
    - properties:
      - Precision

HeikoTheissen: DataHistoryHidden -> DataHistorySnapshot because it is not really hidden, you can retrieve it with $at.
DataHistoryVisible -> DataHistoryFull

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -75,27 +75,35 @@
...
 
-      <ComplexType Name="CollectionKindType" Abstract="true">
-        <Annotation Term="Core.Description" String="Kind of temporal collection" />
+      <ComplexType Name="DataHistory" Abstract="true">
+        <Annotation Term="Core.Description" String="Describes how the data history is represented" />
       </ComplexType>
 
-      <ComplexType Name="CollectionKindSnapshot" BaseType="Temporal.CollectionKindType">
-        <Annotation Term="Core.Description" String="Each OData entity corresponds to a temporal object and by default only shows data of the current time slice" />
+      <ComplexType Name="DataHistorySnapshot " BaseType="Temporal.DataHistory">
```
HeikoTheissen, line 93:
```suggestion
      <ComplexType Name="DataHistorySnapshot" BaseType="Temporal.DataHistory">
```

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -75,27 +75,35 @@
...
+        <Annotation Term="Core.Description" String="Each OData entity represents a snapshot of data at a point in application time that is the same for all entities in the collection" />
+        <Annotation Term="Core.LongDescription">
+          <String>The point in application time is defined by the `$at` clause of a request; default is the current point in time.</String>
+        </Annotation>
       </ComplexType>
 
-      <ComplexType Name="CollectionKindTimeSlice" BaseType="Temporal.CollectionKindType">
-        <Annotation Term="Core.Description" String="Each OData entity corresponds to a time slice and all time slices of a temporal object are part of the collection" />
+      <ComplexType Name="DataHistoryVisible " BaseType="Temporal.DataHistory">
```
HeikoTheissen, line 100:
```suggestion
      <ComplexType Name="DataHistoryVisible" BaseType="Temporal.DataHistory">
```

HeikoTheissen (changes requested):

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,98 +67,87 @@
...
+        </Annotation>
+      </ComplexType>
 
-      <ComplexType Name="GranularityDateTimeOffset" BaseType="Temporal.TimeDimensionType">
-        <Annotation Term="Core.Description" String="Granularity of this time dimension is DateTimeOffset" />
+      <ComplexType Name="DataHistoryVisible" BaseType="Temporal.DataHistory">
+        <Annotation Term="Core.Description" String="Each OData entity represents data during a period of application time; it is called a *time slice*" />
+        <Annotation Term="Core.LongDescription">
+          <String>A temporal collection of this kind MUST NOT contain two entities with the same
+          `ObjectKey` whose application-time defined by their `TimeSliceStart` and `TimeSliceEnd` properties overlap;
```
HeikoTheissen, line 104:
```suggestion
          `ObjectKey` whose application-time period defined by their `TimeSliceStart` and `TimeSliceEnd` properties overlap;
```
Word missing?

HeikoTheissen (changes requested):

HeikoTheissen (approved):

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,98 +67,87 @@
...
         </Property>
-        <Property Name="ObjectKey" Type="Collection(Edm.PropertyPath)" Nullable="false">
-          <Annotation Term="Core.Description" String="The set of primitive properties that identify a temporal object" />
-          <Annotation Term="Core.LongDescription" String="A temporal object is a set of facts whose change over application and/or system time is tracked by the service. The entities of the annotated entity set are time slices of temporal objects. A temporal object is uniquely identified within this set by the values of the specified object key properties. Object key properties follow the same rules as entity key properties. If no object key is specified, the object key is identical to the entity key." />
+        <Property Name="UnitOfTime" Type="Temporal.UnitOfTime" Nullable="false">
+          <Annotation Term="Core.Description" String="Unit of time and other properties of a time period" />
         </Property>
         <Property Name="SupportedActions" Type="Collection(Temporal.QualifiedActionName)" Nullable="false">
           <Annotation Term="Core.Description" String="List of supported temporal actions" />
         </Property>
```
HeikoTheissen, line 86: Move this to `DataHistoryVisible`, because it is only relevant there.

HeikoTheissen (changes requested):

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,98 +67,87 @@
...
       </ComplexType>
 
-      <ComplexType Name="GranularityDate" BaseType="Temporal.TimeDimensionType">
-        <Annotation Term="Core.Description" String="Granularity of this time dimension is Date" />
+      <ComplexType Name="UnitOfTimeDate" BaseType="Temporal.UnitOfTime">
+        <Annotation Term="Core.Description" String="Period start and end are of type Edm.Date" />
         <Property Name="ClosedClosedPeriods" Type="Edm.Boolean" Nullable="false" DefaultValue="false">
-          <Annotation Term="Core.Description" String="Application-time periods are stored with closed-closed semantics" />
-          <Annotation Term="Core.LongDescription" String="Use the temporal query option `$toInclusive` instead of `$to` for specifying the period end in time-series queries." />
+          <Annotation Term="Core.Description" String="If `true`, the period end is included in the time period" />
```
HeikoTheissen, line 133:
```suggestion
          <Annotation Term="Core.Description" String="If `true`, the period end is the last day in the time period; if `false`, the period end is the first day after the period" />
```

HeikoTheissen (commented):
Question: When does the closed-closed date period [2021-06-01, 2021-06-30] end?
- [ ] with the last day of June
- [ ] at 2021-07-01T00:00:00
- [ ] at 2021-06-30T00:00:00

What do you think?

ralfhandl:
> Question: When does the closed-closed date period [2021-06-01, 2021-06-30] end?

It ends at the end of June 30, 2021 😄

HeikoTheissen (approved):

HeikoTheissen (changes requested):
To be in line with other example files, rename `examples/Org.OData.Temporal.V1-history-sample.*` to `examples/Org.OData.Temporal.V1.history-sample.*` and likewise for `snapshot`.

ralfhandl:
The other examples are named <QualifiedTermName>-sample. These are named <Vocabulary>-<topic>-sample. Which is an intentional extension to the original naming scheme.

I could use `examples/Org.OData.Temporal.V1.ApplicationTimeSupport-history-sample.*`. etc.

HeikoTheissen:
> The other examples are named `<QualifiedTermName>-sample.*`.

But `Org.OData.Capabilities.V1.permissions` is not a term, so this would belong into the new category and must be renamed to `Org.OData.Capabilities.V1-permissions-sample.*`.

examples/Org.OData.Temporal.V1-history-sample.xml
```diff
@@ -0,0 +1,104 @@
...
+          </Record>
+        </Annotation>
+      </Annotations>
+
+      <Annotations Target="OrgModel.Default/Departments/history">
+        <Annotation Term="Temporal.ApplicationTimeSupport">
+          <Record>
+            <PropertyValue Property="DataHistory">
+              <Record Type="Temporal.DataHistoryTimeSlice">
+                <PropertyValue Property="TimeSliceStart" PropertyPath="history/From" />
```
HeikoTheissen, line 84: I believe the path evaluation rules for this `TimeSliceStart` are undefined in the spec, see [this comment](https://issues.oasis-open.org/browse/ODATA-1420?focusedCommentId=79346&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-79346).

HeikoTheissen (approved):

HeikoTheissen (commented):
@ralfhandl, is this OK?

ralfhandl (commented):
Nice!

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -231,46 +223,36 @@
...
-        <Parameter Name="From" Type="Edm.PrimitiveType" Nullable="false">
-          <Annotation Term="Core.Description" String="Lower boundary (inclusive) of application time. Date or DateTimeOffset value, or string with value `min`" />
-        </Parameter>
-        <Parameter Name="To" Type="Edm.PrimitiveType" Nullable="false">
-          <Annotation Term="Core.Description" String="Upper boundary (exclusive) of application time. Date or DateTimeOffset value, or string with value `max`" />
-        </Parameter>
-        <Parameter Name="ObjectKey" Type="Edm.EntityType" Nullable="false">
-          <Annotation Term="Core.OptionalParameter">
-            <Record />
+        <Parameter Name="deletionTimeslices" Type="Collection(Edm.EntityType)" Nullable="false">
```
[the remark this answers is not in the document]
HeikoTheissen: "new" sounds as if they contribute something new to the collection. But these actually take something away. If we want unique naming, what about `deltaTimeslices` everywhere?

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,214 +67,192 @@
...
         </Parameter>
         <ReturnType Type="Collection(Edm.EntityType)" Nullable="false">
           <Annotation Term="Core.Description" String="Modified time slices" />
           <Annotation Term="Core.LongDescription" String="Same entity type as binding parameter." />
         </ReturnType>
       </Action>
 
-      <Action Name="SynchronizeObject" IsBound="true" EntitySetPath="timeslices">
+      <Action Name="Synchronize" IsBound="true" EntitySetPath="timeslices">
         <Annotation Term="Core.Description" String="Synchronizes time slices of a temporal object to match the provided full application timeline" />
```
HeikoTheissen, line 208: Unclear what this means. Is `timeslices` simply replaced with `timeline` (per object key)?

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,214 +67,192 @@
...
+          </Annotation>
         </Parameter>
         <ReturnType Type="Collection(Edm.EntityType)" Nullable="false">
           <Annotation Term="Core.Description" String="Modified time slices" />
           <Annotation Term="Core.LongDescription" String="Same entity type as binding parameter." />
         </ReturnType>
       </Action>
 
-      <Action Name="SynchronizeObject" IsBound="true" EntitySetPath="timeslices">
+      <Action Name="Synchronize" IsBound="true" EntitySetPath="timeslices">
```
HeikoTheissen, line 207:
```suggestion
      <Action Name="Synchronize" IsBound="true">
```
No EntitySetPath without ReturnType.

HeikoTheissen (changes requested):

vocabularies/Org.OData.Temporal.V1.xml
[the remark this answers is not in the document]
ralfhandl: Everywhere except for `Synchronize`

vocabularies/Org.OData.Temporal.V1.xml
ralfhandl: The resulting history is semantically equivalent, i.e. the same values at the same point in time, but potentially with a different cut of time slices - we can't require services to merge adjacent and value-identical time slices.

HeikoTheissen (approved):

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -137,6 +136,23 @@
         </Property>
       </ComplexType>
 
+      <ComplexType Name="Period">
+        <Annotation Term="Core.Description" String="A period to be changed with an action defined in this vocabulary" />
+        <Annotation Term="Core.LongDescription">
+          <String>The action parameter with this type MUST be provided if the action is invoked on a collection
+          without visible data history. It MUST NOT be provided otherwise.
+          Either both `TimeSliceStart` and `TimeSliceEnd` are of type `Edm.Date`
+          or both are of type `Edm.DateTimeOffset`,
```
ralfhandl, line 145:
```suggestion
          `PeriodStart` and `PeriodEnd` MUST have the same data type, either `Edm.Date` or `Edm.DateTimeOffset`,
```

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -137,6 +136,23 @@
...
+          <String>The action parameter with this type MUST be provided if the action is invoked on a collection
+          without visible data history. It MUST NOT be provided otherwise.
+          Either both `TimeSliceStart` and `TimeSliceEnd` are of type `Edm.Date`
+          or both are of type `Edm.DateTimeOffset`,
+          depending on the `ApplicationTimeSupport/UnitOfTime` of the collection on which the action is invoked.</String>
+        </Annotation>
+        <Property Name="TimeSliceStart" Type="Edm.PrimitiveType" Nullable="false">
+          <Annotation Term="Core.Description" String="Lower boundary of a time slice" />
+        </Property>
+        <Property Name="TimeSliceEnd" Type="Edm.PrimitiveType">
```
ralfhandl, line 151:
```suggestion
        <Property Name="PeriodEnd" Type="Edm.PrimitiveType" Nullable="true">
```
To avoid later questions on whether this was intentional.

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -137,6 +136,23 @@
...
+      <ComplexType Name="Period">
+        <Annotation Term="Core.Description" String="A period to be changed with an action defined in this vocabulary" />
+        <Annotation Term="Core.LongDescription">
+          <String>The action parameter with this type MUST be provided if the action is invoked on a collection
+          without visible data history. It MUST NOT be provided otherwise.
+          Either both `TimeSliceStart` and `TimeSliceEnd` are of type `Edm.Date`
+          or both are of type `Edm.DateTimeOffset`,
+          depending on the `ApplicationTimeSupport/UnitOfTime` of the collection on which the action is invoked.</String>
+        </Annotation>
+        <Property Name="TimeSliceStart" Type="Edm.PrimitiveType" Nullable="false">
```
ralfhandl, line 148:
```suggestion
        <Property Name="PeriodStart" Type="Edm.PrimitiveType" Nullable="false">
```
We currently use "period start" in the specification.

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -159,6 +175,9 @@
             New time slices are processed in the order of the collection, which especially matters if some of the specified change periods overlap.</String>
           </Annotation>
         </Parameter>
+        <Parameter Name="deltaIntervals" Type="Collection(Temporal.Period)" Nullable="false">
+          <Annotation Term="Core.Description" String="Periods in 1:1 correspondence with `deltaTimeslices` in case the collection to be modified does not have visible data history" />
+        </Parameter>
```
ralfhandl: That is too error prone: better have one input collection with both the period and the data

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -167,7 +166,7 @@
         <Parameter Name="timeslices" Type="Collection(Edm.EntityType)" Nullable="false">
           <Annotation Term="Core.Description" String="Time slices to modify" />
         </Parameter>
-        <Parameter Name="deltaTimeslices" Type="Collection(Edm.EntityType)" Nullable="false">
+        <Parameter Name="deltaTimeslices" Type="Collection(Temporal.TimesliceWithPeriod)" Nullable="false">
           <Annotation Term="Core.Description" String="New time slices whose property values are used to update the `timeslices` collection" />
           <Annotation Term="Core.LongDescription">
             <String>Same entity type as binding parameter.
```
ralfhandl, line 172:
```suggestion
            <String>The `Timeslice` property MUST have the same entity type as the binding parameter.
```
Also in subsequent similar places.

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -137,20 +137,19 @@
...
-      <ComplexType Name="Period">
-        <Annotation Term="Core.Description" String="Validity period of a &quot;delta timeslice&quot; in an action defined in this vocabulary" />
+      <ComplexType Name="TimesliceWithPeriod">
+        <Annotation Term="Core.Description" String="Delta timeslices with validity period in an action defined in this vocabulary" />
         <Annotation Term="Core.LongDescription">
-          <String>The action parameter with this type MUST be provided if the action is invoked on a collection
-          without visible data history. It MUST NOT be provided otherwise.
-          Either both `TimeSliceStart` and `TimeSliceEnd` are of type `Edm.Date`
-          or both are of type `Edm.DateTimeOffset`,
+          <String>The properties `PeriodStart` and `PeriodEnd` MUST NOT be present if the collection has visible data history.
```
ralfhandl, line 143:
```suggestion
          <String>The properties `PeriodStart` and `PeriodEnd` MUST NOT be present if the provided type of `Timeslice` has visible data history.
```
Well, that's not a good proposal: maybe we should be silent here and instead mention this restriction on every use, where we can refer to the collection the action is bound to.

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -137,20 +137,19 @@
...
+           Otherwise, they MUST have the same type, either `Edm.Date` or `Edm.DateTimeOffset`,
           depending on the `ApplicationTimeSupport/UnitOfTime` of the collection on which the action is invoked.</String>
         </Annotation>
-        <Property Name="TimeSliceStart" Type="Edm.PrimitiveType" Nullable="false">
-          <Annotation Term="Core.Description" String="Lower boundary of a time slice" />
+        <Property Name="Timeslice" Type="Edm.EntityType" Nullable="false"/>
```
ralfhandl, line 147:
```suggestion
        <Property Name="Timeslice" Type="Edm.EntityType" Nullable="false">
          <Annotation Term="Core.Description" String="Time slice data" />
        </Property>
```
Not really a good description, but we should have one here.

ralfhandl (commented):
I think we are getting there...

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,214 +68,208 @@
...
+          <Annotation Term="Core.Description" String="List of supported temporal actions" />
+        </Property>
+      </ComplexType>
 
+      <ComplexType Name="TimesliceWithPeriod">
+        <Annotation Term="Core.Description" String="Delta timeslices with validity period in an action defined in this vocabulary" />
+        <Annotation Term="Core.LongDescription">
+          <String>The properties `PeriodStart` and `PeriodEnd` MUST NOT be present
+          if the entity type of the `Timeslice` already contains time slice start and end, that is,
+          if collection on which the action is invoked has visible data history.
```
ralfhandl, line 145:
```suggestion
          if the collection on which the action is invoked has visible data history.
```

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,214 +68,208 @@
...
+        <Property Name="ObjectKey" Type="Collection(Edm.PropertyPath)" Nullable="false">
+          <Annotation Term="Core.Description" String="The set of primitive properties that identify a temporal object" />
+          <Annotation Term="Core.LongDescription" String="A temporal object is a set of facts whose changes over application time are tracked by the service. The entities in the annotated collection belong to potentially multiple temporal objects, and each temporal object is uniquely identified by the values of the specified object key properties. Object key properties follow the same rules as entity key properties. If no object key is specified, only a single temporal object belongs to the annotated collection." />
+        </Property>
+        <Property Name="SupportedActions" Type="Collection(Core.QualifiedActionName)" Nullable="false">
+          <Annotation Term="Core.Description" String="List of supported temporal actions" />
+        </Property>
+      </ComplexType>
 
+      <ComplexType Name="TimesliceWithPeriod">
```
HeikoTheissen, line 140: Alternatively, the period boundaries could be brought into a snapshot entity set as instance annotations `@Temporal.ApplicationTimeSupport/DataHistory/TimeSliceStart` and `TimeSliceEnd` if these were of `Edm.PrimitiveType` rather than `Edm.PropertyPath`.

ralfhandl: Yes, and it would require support of instance annotations in request payloads, which is rare.

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,214 +68,208 @@
...
+          <String>To address an entity, a point in application time must be specified as described in [OData-Temporal, section 4.2.1].</String>
+        </Annotation>
+      </ComplexType>
 
-      <Term Name="History" Type="Collection(Edm.EntityType)" Nullable="false">
-        <Annotation Term="Core.Description" String="A collection of time slices representing the history of the annotated entity" />
-        <Annotation Term="Core.LongDescription" String="The time slices have the same entity type as the annotated entity and use the same select-expand projection." />
-      </Term>
+      <ComplexType Name="DataHistoryVisible" BaseType="Temporal.DataHistory">
+        <Annotation Term="Core.Description" String="Each OData entity represents data during a period of application time; it is called a *time slice*" />
```
HeikoTheissen, line 119:
```suggestion
        <Annotation Term="Core.Description" String="Each OData entity represents data during a period of application time" />
```

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,214 +68,208 @@
...
+          <String>The temporal collection MUST NOT contain two entities with the same object key as defined by their `ObjectKey` properties
+          and with overlapping application-time periods as defined by their `TimeSliceStart` and `TimeSliceEnd` properties.
+          The temporal collection always contains all entities (with consecutive time periods) for a given object key.</String>
+        </Annotation>
+        <Property Name="TimeSliceStart" Type="Edm.PropertyPath" Nullable="false">
+          <Annotation Term="Core.Description" String="Property containing lower boundary of a time slice" />
+        </Property>
+        <Property Name="TimeSliceEnd" Type="Edm.PropertyPath" Nullable="false">
+          <Annotation Term="Core.Description" String="Property containing upper boundary of a time slice" />
+        </Property>
```
HeikoTheissen, line 130:
```suggestion
          <String>The temporal collection MUST NOT contain two entities with the same object key as defined by their `ObjectKey` properties
          and with overlapping application-time periods as defined by their `PeriodStart` and `PeriodEnd` properties.
          The temporal collection always contains all entities (with consecutive time periods) for a given object key.</String>
        </Annotation>
        <Property Name="PeriodStart" Type="Edm.PropertyPath" Nullable="false">
          <Annotation Term="Core.Description" String="Property containing lower boundary of a period" />
        </Property>
        <Property Name="PeriodEnd" Type="Edm.PropertyPath" Nullable="false">
          <Annotation Term="Core.Description" String="Property containing upper boundary of a period" />
        </Property>
```

HeikoTheissen (changes requested):
When speaking of entity sets with visible data history, avoid the term "time slice" in favor of "period". Speak of "time slices" in the action definitions, which can be applied to entity sets with or without visible data history.

HeikoTheissen (approved):

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,214 +68,209 @@
...
+        <Property Name="PeriodEnd" Type="Edm.PropertyPath" Nullable="false">
+          <Annotation Term="Core.Description" String="Property containing upper boundary of a period" />
+        </Property>
+        <Property Name="ObjectKey" Type="Collection(Edm.PropertyPath)" Nullable="false">
+          <Annotation Term="Core.Description" String="The set of primitive properties that identify a temporal object" />
+          <Annotation Term="Core.LongDescription" String="A temporal object is a set of facts whose changes over application time are tracked by the service. The entities in the annotated collection belong to potentially multiple temporal objects, and each temporal object is uniquely identified by the values of the specified object key properties. Object key properties follow the same rules as entity key properties. If no object key is specified, only a single temporal object belongs to the annotated collection." />
+        </Property>
+        <Property Name="SupportedActions" Type="Collection(Core.QualifiedActionName)" Nullable="false">
+          <Annotation Term="Core.Description" String="List of supported temporal actions" />
+        </Property>
```
HeikoTheissen, line 138:
```suggestion
```

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -67,214 +68,209 @@
...
         </Property>
-        <Property Name="ObjectKey" Type="Collection(Edm.PropertyPath)" Nullable="false">
-          <Annotation Term="Core.Description" String="The set of primitive properties that identify a temporal object" />
-          <Annotation Term="Core.LongDescription" String="A temporal object is a set of facts whose change over application and/or system time is tracked by the service. The entities of the annotated entity set are time slices of temporal objects. A temporal object is uniquely identified within this set by the values of the specified object key properties. Object key properties follow the same rules as entity key properties. If no object key is specified, the object key is identical to the entity key." />
-        </Property>
-        <Property Name="SupportedActions" Type="Collection(Temporal.QualifiedActionName)" Nullable="false">
-          <Annotation Term="Core.Description" String="List of supported temporal actions" />
+        <Property Name="DataHistory" Type="Temporal.DataHistory" Nullable="false">
+          <Annotation Term="Core.Description" String="Describes how the data history is represented" />
         </Property>
```
HeikoTheissen, line 84:
```suggestion
        <Property Name="DataHistory" Type="Temporal.DataHistory" Nullable="false">
          <Annotation Term="Core.Description" String="Describes how the data history is represented" />
        </Property>
        <Property Name="SupportedActions" Type="Collection(Core.QualifiedActionName)" Nullable="false">
          <Annotation Term="Core.Description" String="List of supported temporal actions" />
        </Property>
```

HeikoTheissen (changes requested):
SupportedActions independent of the DataHistory

HeikoTheissen (approved):

PR #128, opened 2021-06-11
TC homework 2021-06-10

HeikoTheissen:


vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -95,11 +95,22 @@
...
       </ComplexType>
 
       <ComplexType Name="CollectionKindTimeSlice" BaseType="Temporal.CollectionKindType">
-        <Annotation Term="Core.Description" String="Each OData entity corresponds to a time slice and all time slices of a temporal object are part of the collection" />
+        <Annotation Term="Core.Description" String="Each OData entity represents data during an interval of application time; it is called a *time slice*" />
+        <Annotation Term="Core.LongDescription">
+          <String>An entity in a temporal collection of this kind and an instance of the same entity type are said to *overlap*
+          if the intervals defined by their `TimeSliceStart` and `TimeSliceEnd` properties overlap
+          and their other ("non-temporal") key properties are identical.
+          A temporal collection of this kind MUST NOT contain overlapping entities from the same entity set;
```
HeikoTheissen, line 110: Is this statement correct, @ralfhandl?

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -136,30 +147,17 @@
       <!-- Operations on Temporal Objects -->
 
       <Action Name="Update" IsBound="true" EntitySetPath="timeslices">
-        <Annotation Term="Core.Description" String="Updates all time slices of a temporal object in a given period of application time after splitting partially overlapping time slices" />
+        <Annotation Term="Core.Description" String="Adds time slices to a time-slice entity set, splitting other time slices in case of overlap" />
```
ralfhandl, line 150: Some of the splitted time slices are updated, which was part of the original text.

Let's first try to describe this in the specification, then copy parts of the text here.

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -136,30 +147,17 @@
...
-        <Parameter Name="From" Type="Edm.PrimitiveType" Nullable="false">
-          <Annotation Term="Core.Description" String="Lower boundary (inclusive) of application time. Date or DateTimeOffset value, or string with value `min`" />
-        </Parameter>
-        <Parameter Name="To" Type="Edm.PrimitiveType" Nullable="false">
-          <Annotation Term="Core.Description" String="Upper boundary (exclusive) of application time. Date or DateTimeOffset value, or string with value `max`" />
-        </Parameter>
-        <Parameter Name="Values" Type="Edm.EntityType" Nullable="false">
-          <Annotation Term="Core.Description" String="Property values to update" />
-          <Annotation Term="Core.LongDescription" String="Same entity type as binding parameter and return type, specify only values of non-key properties." />
+          <Annotation Term="Core.Description" String="Collection of time slices to which new time slices are added" />
```
ralfhandl, line 152: It's mostly update, and insert only as a result of splitting. 

Again, better get the prose specification right, then copy text here.

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -136,30 +147,17 @@
...
         </Parameter>
-        <Parameter Name="ObjectKey" Type="Edm.EntityType" Nullable="false">
-          <Annotation Term="Core.OptionalParameter">
-            <Record />
-          </Annotation>
-          <Annotation Term="Core.Description" String="Property values that identify a single temporal object" />
-          <Annotation Term="Core.LongDescription" String="Select-expand projection of the same entity type as binding parameter, containing all and only properties reference in the [`ObjectKey`](#ObjectKey) of that entity type." />
+        <Parameter Name="newTimeslices" Type="Collection(Edm.EntityType)" Nullable="false">
```
ralfhandl, line 154: Actually these are not full and new time slices, it is a single partial time slice, consisting of the interval and the values to change during that interval:

```suggestion
        <Parameter Name="IntervalAndValuesToUpdate" Type="Edm.EntityType" Nullable="false">
```

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -136,30 +147,17 @@
...
+          <Annotation Term="Core.Description" String="Time slices to be added into the collection" />
+          <Annotation Term="Core.LongDescription" String="Same entity type as binding parameter" />
         </Parameter>
         <ReturnType Type="Collection(Edm.EntityType)" Nullable="false">
           <Annotation Term="Core.Description" String="Modified time slices" />
-          <Annotation Term="Core.LongDescription" String="Same entity type as binding parameter." />
+          <Annotation Term="Core.LongDescription" String="Same entity type as binding parameter" />
```
ralfhandl, line 160:
```suggestion
          <Annotation Term="Core.LongDescription" String="Same entity type as binding parameter." />
```

We usually have a period at the end of a LongDescription.

ralfhandl (changes requested):

vocabularies/Org.OData.Temporal.V1.xml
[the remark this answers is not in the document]
HeikoTheissen: This is necessary because the entities `EntitySet('A')/ContainmentNavProp(1)` and `EntitySet('B')/ContainmentNavProp(1)` have the same key property `1` and still are distinct because they live in different implicit entity sets.

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -95,11 +95,22 @@
...
       <ComplexType Name="CollectionKindTimeSlice" BaseType="Temporal.CollectionKindType">
-        <Annotation Term="Core.Description" String="Each OData entity corresponds to a time slice and all time slices of a temporal object are part of the collection" />
+        <Annotation Term="Core.Description" String="Each OData entity represents data during an interval of application time; it is called a *time slice*" />
+        <Annotation Term="Core.LongDescription">
+          <String>An entity in a temporal collection of this kind and an instance of the same entity type are said to *overlap*
+          if the intervals defined by their `TimeSliceStart` and `TimeSliceEnd` properties overlap
+          and their other ("non-temporal") key properties are identical.
+          A temporal collection of this kind MUST NOT contain overlapping entities from the same entity set;
+          and if it contains an entity, it MUST contain all entities from the same entity set
+          that agree with the given one in their non-temporal key properties.</String>
```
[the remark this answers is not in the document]
HeikoTheissen: I mention the "same entity set" only because of the containment case (see my comment above).

vocabularies/Org.OData.Temporal.V1.xml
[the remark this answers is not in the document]
HeikoTheissen: The part after the semicolon replaces the following:
https://github.com/oasis-tcs/odata-vocabularies/blob/e6f7b53c199d88f185cb4e375bc8167ca8323508/vocabularies/Org.OData.Temporal.V1.xml#L102

vocabularies/Org.OData.Temporal.V1.xml
HeikoTheissen: Why don't you want to allow a mass update?

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -136,26 +147,16 @@
...
-          <Annotation Term="Core.LongDescription" String="Same entity type as binding parameter and return type, specify only values of non-key properties." />
-        </Parameter>
-        <Parameter Name="ObjectKey" Type="Edm.EntityType" Nullable="false">
-          <Annotation Term="Core.OptionalParameter">
-            <Record />
-          </Annotation>
-          <Annotation Term="Core.Description" String="Property values that identify a single temporal object" />
-          <Annotation Term="Core.LongDescription" String="Select-expand projection of the same entity type as binding parameter, containing all and only properties reference in the [`ObjectKey`](#ObjectKey) of that entity type." />
+        <Parameter Name="newValues" Type="Collection(Edm.EntityType)" Nullable="false">
+          <Annotation Term="Core.Description" String="Non-overlapping time slices whose property values are used to update the `timeslices` collection" />
```
HeikoTheissen, line 158: Is it OK to update values in several intervals with one action call? If so, these intervals must not overlap.

ralfhandl: I think it is ok, and the intervals may even overlap as long as the changes are processed in the order of the "change instructions".

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -136,26 +147,20 @@
...
-          <Annotation Term="Core.LongDescription" String="Same entity type as binding parameter and return type, specify only values of non-key properties." />
-        </Parameter>
-        <Parameter Name="ObjectKey" Type="Edm.EntityType" Nullable="false">
-          <Annotation Term="Core.OptionalParameter">
-            <Record />
+        <Parameter Name="newValues" Type="Collection(Edm.EntityType)" Nullable="false">
+          <Annotation Term="Core.Description" String="Partial time slices whose property values are used to update the `timeslices` collection" />
+          <Annotation Term="Core.LongDescription">
+            <String>Same entity type as binding parameter.
+            The partial time slices need not contain all properties, but at least the boundary values of the interval to change.
```
HeikoTheissen, line 161: If the time-slice collection has non-temporal key properties, but these are not given in `newValues`, all "temporal objects" in the collection will be updated at once. OK?

ralfhandl: Yes, similar to SQL's `UPDATE ... FOR PORTION OF BUSINESS_TIME ... WHERE ... `, see https://www.ibm.com/docs/en/db2oc?topic=statements-update

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -159,7 +159,8 @@
           <Annotation Term="Core.LongDescription">
             <String>Same entity type as binding parameter.
             The partial time slices need not contain all properties, but at least the boundary values of the interval to change.
-            They are processed in the order of the collection, which especially matters if some of the specified change intervals overlap.</String>
+            An absent non-temporal key property overlaps any key property value.
```
ralfhandl, line 162:
```suggestion
            An absent non-temporal key property matches any key property value.
```

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -95,11 +95,22 @@
...
+        </Annotation>
       </ComplexType>
 
       <ComplexType Name="CollectionKindTimeSlice" BaseType="Temporal.CollectionKindType">
-        <Annotation Term="Core.Description" String="Each OData entity corresponds to a time slice and all time slices of a temporal object are part of the collection" />
+        <Annotation Term="Core.Description" String="Each OData entity represents data during an interval of application time; it is called a *time slice*" />
+        <Annotation Term="Core.LongDescription">
+          <String>Entities in a temporal collection of this kind are said to *overlap*
+          if the intervals defined by their `TimeSliceStart` and `TimeSliceEnd` properties overlap
+          and their other, *non-temporal*, key properties are identical.
```
HeikoTheissen, line 109: Are `TimeSliceStart` and `TimeSliceEnd` always key properties (in the OData sense), or can they also be part of a "logical key"? That would complicate the definition of "overlapping".

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -95,11 +95,22 @@
...
+          A temporal collection of this kind MUST NOT contain overlapping entities;
+          and it contains with every entity all entities
+          that agree with the given one in their non-temporal key properties.</String>
+        </Annotation>
         <Property Name="TimeSliceStart" Type="Edm.PropertyPath" Nullable="false">
```
HeikoTheissen, line 114:
```suggestion
        <Property Name="NonTemporalKey" Type="Collection(Edm.PropertyPath)" Nullable="false">
          <Annotation Term="Core.Description" String="Properties that, together with the boundary properties, uniquely identify an entity in the time-slice collection"/>
          <Annotation Term="Core.LongDescription" String="Need not be specified if the boundary properties are key properties."/>
        </Property>
        <Property Name="TimeSliceStart" Type="Edm.PropertyPath" Nullable="false">
```
We could be explicit, of course.

vocabularies/Org.OData.Temporal.V1.xml
ralfhandl: They can be non-key. "Overlapping" should only be used for the time period, not for the "object key"

HeikoTheissen: > "Overlapping" should only be used for the time period, not for the "object key"

A time-slice collection may contain overlapping intervals if the object keys are different. That's why I define "overlapping entities" by looking also at the object key. But since the object key can also be non-key, we must make it explicit (see https://github.com/oasis-tcs/odata-vocabularies/pull/128#discussion_r650726899).

HeikoTheissen:
@ralfhandl: Can temporal actions also be invoked on snapshot collections? If yes, then the proposed changes don't work, because the `newTimeslices` parameter wouldn't contain the period boundaries.

ralfhandl:
@HeikoTheissen I think that is ok. If someone wants their client to change historic data without being able to efficiently read it, then they can define their own actions based on ours, with explicit From and To parameters.

PR #129, opened 2021-06-14
ODATA-1445 - new - Extend AppliesTo of JSON.Schema

ralfhandl:


vocabularies/Org.OData.JSON.V1.xml
```diff
@@ -64,7 +64,7 @@
         </Collection>
       </Annotation>
 
-      <Term Name="Schema" Type="JSON.JSON" Nullable="false" AppliesTo="Property">
+      <Term Name="Schema" Type="JSON.JSON" Nullable="false" AppliesTo="Parameter Property ReturnType Term TypeDefinition">
```
HeikoTheissen, line 67: Can a term have type `Edm.Stream`? [OData-CSDL, sec. 14.3](https://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html#sec_ConstantExpression) does not tell how they can be specified.

ralfhandl: Yes, there is no restriction on the types a term can have, and Edm.Stream is just a primitive type.

See https://issues.oasis-open.org/browse/ODATA-1323 how to specify a stream value in an annotation.

HeikoTheissen: The solution for ODATA-1323 does not propose a new subsection of OData-CSDL, sec. 14.3. Isn't that warranted?

ralfhandl: Good point, added as a comment to ODATA-1323.

HeikoTheissen (approved):

vocabularies/Org.OData.JSON.V1.xml
GeraldKrause: I would rather add it to the proposal text. Based on my experiences there are chances that it will get lost otherwise.

GeraldKrause (commented):
I would rather add it to the proposal text. Based on my experiences there are chances that it will get lost otherwise.

vocabularies/Org.OData.JSON.V1.xml
ralfhandl: Done

vocabularies/Org.OData.JSON.V1.xml
```diff
@@ -66,7 +66,7 @@
 
       <Term Name="Schema" Type="JSON.JSON" Nullable="false" AppliesTo="Parameter Property ReturnType Term TypeDefinition">
         <Annotation Term="Core.RequiresType" String="Edm.Stream" />
-        <Annotation Term="Core.Description" String="The JSON Schema for this JSON property" />
+        <Annotation Term="Core.Description" String="The JSON Schema for JSON values the annotated property, parameter, return type, term, or type definition" />
```
HeikoTheissen, line 69:
```suggestion
        <Annotation Term="Core.Description" String="The JSON Schema for JSON values of the annotated property, parameter, return type, term, or type definition" />
```
"of" missing?

HeikoTheissen (changes requested):

HeikoTheissen (approved):

PR #134, opened 2021-07-08
Avoid dependency on name of main branch

HeikoTheissen:
The external type links in the vocabulary markdown files refer to the oasis-tcs/odata-vocabularies repository and contain the name of the main branch (currently `master`). Suggestion to replace them with links to github.io.

### DIFF
--- a/test/csdl2markdown.test.js
+++ b/test/csdl2markdown.test.js
@@ -96,7 +96,7 @@ describe("Non-OASIS Vocabularies", function () {
       "",
       "Term|Type|Description",
       ":---|:---|:----------",
-      'Foo *(Deprecated)*|[Tag](https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Core.V1.md#Tag)|<a name="Foo"></a>Use term `Foo` from the OASIS Core vocabulary instead',
+      'Foo *(Deprecated)*|[Tag](https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Core.V1.html#Tag)|<a name="Foo"></a>Use term `Foo` from the OASIS Core vocabulary instead',
       'Bar|[Type](Another.md#Type)|<a name="Bar"></a>',
       "",
     ];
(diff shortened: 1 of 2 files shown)


test/csdl2markdown.test.js
```diff
@@ -96,7 +96,7 @@ describe("Non-OASIS Vocabularies", function () {
       "",
       "Term|Type|Description",
       ":---|:---|:----------",
-      'Foo *(Deprecated)*|[Tag](https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Core.V1.md#Tag)|<a name="Foo"></a>Use term `Foo` from the OASIS Core vocabulary instead',
+      'Foo *(Deprecated)*|[Tag](https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Core.V1.md#Tag)|<a name="Foo"></a>Use term `Foo` from the OASIS Core vocabulary instead',
```
ralfhandl, line 99: Try following this link and you'll understand why we do not link there.

ralfhandl (changes requested):
We intentionally link to the *.md files via the GitHub UI, which displays them nicely, not as raw markdown.

Unfortunately I haven't found a way to do that without mentioning the branch name.

Let's rename the main branch to "main" first and then replace all necessary occurrences of "master".

test/csdl2markdown.test.js
```diff
@@ -96,7 +96,7 @@ describe("Non-OASIS Vocabularies", function () {
       "",
       "Term|Type|Description",
       ":---|:---|:----------",
-      'Foo *(Deprecated)*|[Tag](https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Core.V1.md#Tag)|<a name="Foo"></a>Use term `Foo` from the OASIS Core vocabulary instead',
+      'Foo *(Deprecated)*|[Tag](https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Core.V1.html#Tag)|<a name="Foo"></a>Use term `Foo` from the OASIS Core vocabulary instead',
```
ralfhandl, line 99: The problem with the HTML version is that links to the XML source code position do not work, this only works from Markdown files viewed via the GitHub UI.

ralfhandl (changes requested):
Links to XML source won't work any more with this change.

PR #135, opened 2021-07-08
Prepare renaming "master" to "main"

ralfhandl:


HeikoTheissen (changes requested):
https://github.com/oasis-tcs/odata-vocabularies/blob/024b57699d057db6bcc3abb58f47326e302f3d50/lib/csdl2markdown.js#L664
Change this as well.

ralfhandl:
@HeikoTheissen are you sure that `HEAD` always points to the tip of the default branch?

In git it points to the tip of the checked-out branch, and I can't find Github documentation on `HEAD`.

HeikoTheissen (approved):

PR #136, opened 2021-07-08
Chore: cleanup after renaming master to main

ralfhandl:


lib/csdl2markdown.js
```diff
@@ -723,7 +728,7 @@ module.exports.csdl2markdown = function (filename, csdl, referenced = {}) {
         "IsURL",
         "LongDescription",
         "Revisions",
-        "OptionalParameter"
+        "OptionalParameter",
```
HeikoTheissen, line 731:
```suggestion
        "OptionalParameter"
```

HeikoTheissen (changes requested):

PR #139, opened 2021-07-16
text fragments in sourceLink

HeikoTheissen:
The vocabulary markdown pages contain links to the XML source, these automatically highlight a certain model element in github.com, e.g., https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Aggregation.V1.xml#L71

But the highlighting does not work in github.io, because the "naked" XML has no line-counting mechanism: https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Aggregation.V1.xml#L71

This pull request adds "text fragments" to the links, which cause (modern) browsers to highlight the model element name, and this works in both cases:
https://github.com/oasis-tcs/odata-vocabularies/blob/text-fragments/vocabularies/Org.OData.Aggregation.V1.xml#L71:~:text=Name=%22-,ApplySupported,-%22
https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Aggregation.V1.xml#L71:~:text=Name=%22-,ApplySupported,-%22


ralfhandl (changes requested):
This breaks the links from term name to term definition in the XML - maybe some quoting is necessary for (one of) the special characters in the fragment?

HeikoTheissen:
> This breaks the links from term name to term definition in the XML - maybe some quoting is necessary for (one of) the special characters in the fragment?

A link is generated if I escape the colon as `%3A`, but then the text fragment is no longer recognized by the browser.

HeikoTheissen:
I found a way to make the text fragment links work in all cases.

ralfhandl (approved):

chrisspre:
Looks good to me. Great feature. Do you have a link to the documentation of this browser feature

HeikoTheissen:
> Looks good to me. Great feature. Do you have a link to the documentation of this browser feature

https://wicg.github.io/scroll-to-text-fragment/

PR #141, opened 2021-07-23
ODATA-1461 - new - Term describing GeoJSON Feature Objects

ralfhandl:
Resolves https://issues.oasis-open.org/browse/ODATA-1461


vocabularies/Org.OData.Core.V1.xml
```diff
@@ -525,6 +525,22 @@ Any simple identifier | Any type listed in `Validation.OpenPropertyTypeConstrain
         <Annotation Term="Validation.Pattern" String="^[\p{L}\p{Nl}_][\p{L}\p{Nl}\p{Nd}\p{Mn}\p{Mc}\p{Pc}\p{Cf}]{0,}$" />
       </TypeDefinition>
 
+      <Term Name="GeometryFeature" Type="Core.GeometryFeatureType">
```
ralfhandl, line 528: Should the annotation target a Geo property instead of a structured type? In this case the `geometry` property would not be needed and be implied by the annotation target.

PR #145, opened 2021-08-09
ODATA-1466 - new - Custom aggregation methods as functions

HeikoTheissen:
https://issues.oasis-open.org/browse/ODATA-1466


vocabularies/Org.OData.Aggregation.V1.xml
```diff
@@ -127,6 +127,16 @@
         </Property>
       </ComplexType>
 
+      <Term Name="IsCustomAggregationMethod" Type="Core.Tag" DefaultValue="true" Nullable="false" AppliesTo="Function">
+        <Annotation Term="Core.Description" String="The annotated function is a custom aggregation method [OData-Aggr, section 3.1.3.6]"/>
+        <Annotation Term="Core.LongDescription">
+          <String>The function is unbound and has one parameter whose type is a collection of a primitive type, this represents the aggregated property in the input set.
```
ralfhandl, line 133:
```suggestion
          <String>The function is unbound and has one parameter whose type is a collection of a primitive type, this represents the property values in the input set which are to be aggregated.
```

HeikoTheissen:
ODATA-1466 deferred.

PR #152, opened 2021-09-08
Text fragments only for the first overload per function/action

HeikoTheissen:


lib/csdl2markdown.js
```diff
@@ -211,7 +211,10 @@ module.exports.csdl2markdown = function (filename, csdl, referenced = {}) {
     lines.push("");
     lines.push("## " + kind + "s");
 
+    var over = {};
```
ralfhandl, line 214:
```suggestion
    const over = {};
```
Don't use `var` inside functions, use either `const` or `let`.

ralfhandl (changes requested):
Please add a test case for the new code

ralfhandl (approved):
Thanks!

Just noted that we do not cover https://github.com/oasis-tcs/odata-vocabularies/blob/c1d7b93d335f20fb8e7b1f9c5444afb0f7e7d699/lib/csdl2markdown.js#L619 in the tests - we do cover it via the Validation vocabulary, so we'll notice when this breaks.

ralfhandl (approved):

PR #153, opened 2021-09-14
Core.AppliesViaContainer shall not implicitly extend AppliesTo

HeikoTheissen:
CHANGELOG
- Rephrase the concession which allows additionally annotating entity types (has been misunderstood by some)
- Terms which make use of the concession must explicitly `AppliesTo="EntityType ..."`


ralfhandl (approved):

ralfhandl (approved):

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -68,7 +68,7 @@
       </Annotation>
 
 
-      <Term Name="ApplicationTimeSupport" Type="Temporal.ApplicationTimeSupportType" Nullable="false" AppliesTo="Collection">
+      <Term Name="ApplicationTimeSupport" Type="Temporal.ApplicationTimeSupportType" Nullable="false" AppliesTo="Collection EntityType">
```
ralfhandl, line 71: Why should we allow `EntityType` here?

ralfhandl (changes requested):

vocabularies/Org.OData.Temporal.V1.xml
HeikoTheissen: At the moment, annotation if entity types is _implicitly_ allowed by `Core.AppliesViaContainer`. We wanted to have this explicit (in order to satisfy the CAP compiler).

Shall we exempt the temporal vocabulary from this rule?

ralfhandl: Yes, please revert the change to the Temporal vocabulary

ralfhandl (approved):

PR #154, opened 2021-09-15
 Temporal: example for timeline with object key

ralfhandl:


HeikoTheissen (approved):

examples/Org.OData.Temporal.V1.objectkey-sample.xml
```diff
@@ -0,0 +1,64 @@
...
+    <edmx:Include Alias="Temporal" Namespace="Org.OData.Temporal.V1" />
+  </edmx:Reference>
+  <edmx:DataServices>
+    <Schema xmlns="http://docs.oasis-open.org/odata/ns/edm" Alias="this" Namespace="org.example.odata.costcenter">
+
+      <EntityType Name="CostCenter">
+        <Key>
+          <PropertyRef Name="AreaID" />
+          <PropertyRef Name="ID" />
+          <PropertyRef Name="ValidTo" />
```
bernhardthimmel, line 16: Usually it would be more common to use ValidFrom as key component since inserting new later time slices would change the ValidTo but not the ValidFrom of the existing record.

ralfhandl: Agreed. Here I wanted to use a "not so common" pattern that is actually used, see https://api.sap.com/api/API_COSTCENTER_SRV/resource

PR #156, opened 2021-09-22
Dump git-log to change-log

HeikoTheissen:


HeikoTheissen:
How can I include `npm run changelog` in `nodejs.yml` so that it is executed only upon merge, not for every commit to the PR branch?

ralfhandl:
> How can I include `npm run changelog` in `nodejs.yml` so that it is executed only upon merge, not for every commit to the PR branch?

Try a [conditional step](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsif) with condition
~~~
if: github.event.pull_request.merged == true
~~~

HeikoTheissen:
OK to have CHANGELOG.md in a separate branch without history? Or does this pose problems on github.io? If we generate it via workflow in the same branch, that gives an extra commit each time.

package.json
```diff
@@ -26,6 +26,7 @@
   },
   "scripts": {
     "build": "node lib/transform.js",
+    "changelog": "./tools/changelog.sh",
```
ralfhandl, line 29: Should move to `lib` because we only collect outdated stuff in `tools`

tools/changelog.sh
```diff
@@ -0,0 +1,7 @@
+folder=$(mktemp -d)
+git log | node tools/changelog >$folder/CHANGELOG.md
```
ralfhandl, line 3: `changelog.js` should move to `lib` because only outdated stuff is kept in `tools` === "garbage bin"

tools/changelog.js
```diff
@@ -0,0 +1,54 @@
...
+	})
+	.on("line", function(line) {
+		if (line.startsWith("commit")) {
+			if (entry && entry.length > 2) this.push(entry.join(""));
+			entry = undefined;
+		} else if (/^Date:\s+(.*)$/.test(line))
+			entry = ["# " + RegExp.$1 + "\n"];
+		else if (entry && /^\s+\S/.test(line)) {
+			if (entry.length === 1)
+				entry.push("");
```
ralfhandl, line 41: Will that ignore the first line with non-whitespace text? Why?

A test case would be useful: move script to `lib` and add a mocha test file in `test`.

ralfhandl (changes requested):

tools/changelog.js
HeikoTheissen: The first line contains the name and number of the pull request. Is that useful in the changelog?

ralfhandl: No. A mocha test for `changelog.js` would really be helpful.

test/changelog.test.js
```diff
@@ -0,0 +1,24 @@
+const assert = require("assert");
+const {spawn} = require("child_process");
+const fs = require("fs");
+const transform = require("../lib/changelog.js");
```
ralfhandl, line 4: You currently aren't using this - do you want to keep it to check that this file exists?

test/gitlog.txt
```diff
@@ -0,0 +1,25 @@
...
+
+    Temporal vocabulary (#146)
+
+commit 122dff83ef4f139e3361e5af557519c898ee81bd
+Author: Ralf Handl <ralf.handl@sap.com>
+Date:   Thu Sep 16 17:19:02 2021 +0200
+
+     Temporal: example for timeline with object key (#154)
+    
+    CHANGELOG
```
ralfhandl, line 23: It would be nice to have a second changelog entry to see that both appear.

ralfhandl (commented):
I kind of like this, the only caveat is that we have to be very careful when merging a pull request because we can't edit the commit message later.

test/changelog.test.js
HeikoTheissen: This is called by mocha, so it is used.

ralfhandl: The tests still run if this line is commented out or removed, and ESLINT complains if it is left in.

HeikoTheissen: You meant the `transform`, I deleted that.

test/gitlog.txt
```diff
@@ -24,12 +24,21 @@ Date:   Thu Sep 16 17:24:48 2021 +0200
...
 
-commit 122dff83ef4f139e3361e5af557519c898ee81bd
+commit 122dff93ef4f139e3361e5af557519c898ee81bd
 Author: Ralf Handl <ralf.handl@sap.com>
 Date:   Thu Sep 16 17:19:01 2021 +0200
 
      Temporal: example for timeline with object key (#154)
     
-    CHANGELOG
+    CHANGELOG 122dff73ef4f139e3361e5af557519c898ee81bd
```
ralfhandl, line 33: I think we need a short handbook/README.md explaining this magic

HeikoTheissen: See https://github.com/oasis-tcs/odata-vocabularies/pull/156#issue-740479921

ralfhandl: Will someone find this pull request if they wonder about the magic?

Better add a section to `lib/README.md` that describes this cool mechanism.

HeikoTheissen:
TC 2021-09-30: Investigate whether the commit texts can automatically be filled from the comment at the top of the pull request. Check the format (presence of `CHANGELOG:`) as a prerequisite for merge.

PR #157, opened 2021-09-27
JSON: function "query"

ralfhandl:
Resolves https://issues.oasis-open.org/browse/ODATA-1336

Check which syntax elements are supported by which implementation
- [x] Node.js implementation: https://www.npmjs.com/package/jsonpath (all)
- [x] Java implementation: https://github.com/json-path/JsonPath (all)
- [x] SAP HANA (root, member, single index, wildcard, index range)
- [x] Microsoft SQL Server (root, member, single index)
- [x] Google BigQuery (root, member, single index)


HeikoTheissen (approved):

vocabularies/Org.OData.JSON.V1.xml
```diff
@@ -72,6 +72,53 @@
...
       </Term>
 
+      <Function Name="query">
+        <Annotation Term="Core.Description" String="Query stream values of media type `application/json`, returning a stream value of media type `application/json`" />
+        <Annotation Term="Core.LongDescription">
+          <String>Extracts a JSON value, such as an array, object, or a JSON scalar value (string, number, boolean, or `null`) from the `input` JSON value:
+- If `path` only consists of child member and single subscript operators, it returns the identified single node within `input`, or `null` if no node is identified. 
+- If `path` potentially identifies multiple nodes within `input` (by using recursive descendant, wildcard, array subset, or array filter expressions), it returns an array containing the identified nodes, or an empty array if no node is identified. 
+- If `input` is not a valid JSON value, the function returns `null`.
+- If `path` is `null`, not a valid [JSONPath expression](#Path), or does not match the structure of `input`, the function returns `null`.
```
ralfhandl, line 82: According to the [JSONPath Processing Model](https://datatracker.ietf.org/doc/html/draft-ietf-jsonpath-base-01#section-3.2) the result of executing a valid query is always a list of nodes, even `$` produces a list with the input as its only item. Which makes sense, because even simple property access `$.foo` can produce a list with more than one element because JSON allows duplicate member names, only I-JSON forbids it.

We could follow this in two ways:
- the result of `JSON.query` is always a single JSON text that contains an array
- the result of `JSON.query` is always an OData collection of JSON texts, and the function can be suffixed with an index segment to retrieve only a single JSON text: `JSON.query(...)/0`

HeikoTheissen: The JSON function in Javascript does not support duplicate member names.
```
> JSON.parse('{"a":1,"a":2}')
{ a: 2 }
```
Having to dive into a JSON array or an OData collection even for the simplest queries is too cumbersome for me. I would at least suggest a convenience function `JSON.querySingle` that returns a single object (or null if that is not possible).

vocabularies/Org.OData.JSON.V1.xml
```diff
@@ -68,10 +132,57 @@
...
+        </Parameter>
+        <Parameter Name="path" Type="JSON.Path">
+          <Annotation Term="Core.Description" String="JSONPath expression to be applied to value of `expr`" />
+        </Parameter>
+        <ReturnType Type="JSON.JSON">
+          <Annotation Term="Core.Description" String="JSON value resulting from applying `path` to `input`" />
+        </ReturnType>
+      </Function>
+
+      <Function Name="value" IsBound="true" IsComposable="true">
```
HeikoTheissen, line 160: How can a function returning a primitive type be composable?

HeikoTheissen (changes requested):

vocabularies/Org.OData.JSON.V1.xml
ralfhandl: Good catch!

HeikoTheissen (approved):

HeikoTheissen (approved):

vocabularies/Org.OData.JSON.V1.xml
```diff
@@ -68,10 +132,57 @@
...
+          <String>Extracts a single OData primitive value from the `input` JSON value:
+- If `path` only consists of the root identifier followed by name and index selectors and identifies a single scalar JSON value (string, number, boolean, or `null`) within `input`, it returns the identified single value cast to an OData primitive value (see below).
+- If `path` identifies multiple nodes within `input` (by using descendant, wildcard, union, array subset, or filter selectors), identifies an object or array, or does not identify any node, the function returns `null`.
+- If `input` is not a valid JSON value, the function returns `null`.
+- If `path` is `null`, not a valid [JSONPath expression](#Path), or does not match the structure of `input` (for example applying an index selector to a scalar value), the function returns `null`.
+
+If a single non-null scalar JSON value is identified by `path` within `input`, the function returns that value as an OData primitive value of type
+- `Edm.String` if the value is a JSON string
+- `Edm.Boolean` if the value is `true` or `false`
+- `Edm.Decimal` with unspecified precision and floating scale if the value is a JSON number
```
ralfhandl, line 172: [RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259#section-6) and [RFC 7493](https://datatracker.ietf.org/doc/html/rfc7493#section-2.2) both recommend only sending number values that fit into IEEE 754 binary64 (double precision) numbers, we could use `Edm.Double` here.

HeikoTheissen (approved):

HeikoTheissen (approved):

HeikoTheissen (approved):

sthagen (approved):
LGTM.

vocabularies/Org.OData.JSON.V1.md
```diff
@@ -1,17 +1,166 @@
...
 
 Textual data of media type `application/json`
+
+<a name="Path"></a>
+## [Path](./Org.OData.JSON.V1.xml#L197:~:text=<TypeDefinition%20Name="-,Path,-")
+**Type:** String
+
+[JSONPath](https://datatracker.ietf.org/doc/html/rfc9535) expression
+
+Implementations SHOULD support at least the following subset of JSONPath:
```
sthagen, line 133: I would love to see MUST/SHOULD instead of SHOULD/MAY (here/line 141).

The use of MAY in "offerings" is in my understanding watering the danger that normally MAY indicates. As in
>The Server MAY respond with random payloads on Tuesdays.

ralfhandl: Changed the SHOULD to MUST, because less is nothing 😄 

Kept the MAY, because these are really optional.

sthagen (approved):
Looks even better to me now, thanks.

HeikoTheissen (approved):

HeikoTheissen:
The return type of an SQL function must be known before its execution, therefore `JSON_VALUE` expects a `returning <type>` clause or returns a `varchar`. See the following results of a Javascript SQL client:
```
> SELECT JSON_VALUE('{"a":"10"}', '$.a') AS "v" FROM DUMMY
[ { v: '10' } ]
> SELECT JSON_VALUE('{"a":10}', '$.a') AS "v" FROM DUMMY;
[ { v: '10' } ]
> SELECT JSON_VALUE('{"a":10}', '$.a' RETURNING INTEGER) AS "v" FROM DUMMY;
[ { v: 10 } ]
```

The proposed OData function `JSON.value` can instead be implemented with the SQL function `JSON_QUERY`, which returns a stringified JSON:
JSON|`SELECT JSON_QUERY('...', '$.a') AS "v" FROM DUMMY`|`JSON.v(path='$.a')`
---|---|---
`{"a":"10"}`|`[ { v: '"10"' } ]`|String `"10"`
`{"a":10}`|`[ { v: '10' } ]`|Number `10`
`{"a":true}`|`[ { v: 'true' } ]`|Boolean `true`

vocabularies/Org.OData.JSON.V1.xml
```diff
@@ -14,64 +14,202 @@
...
+      <Function Name="query" IsBound="true" IsComposable="true">
+        <Annotation Term="Core.Description"
+          String="Query a stream value of media type `application/json`, returning a stream value of media type `application/json`" />
+        <Annotation Term="Core.LongDescription">
+          <String>Extracts a JSON value, such as an array, object, or a JSON scalar value (string, number, boolean, or `null`) from the `input` JSON value:
+- If `path` only consists of the root identifier followed by name and index selectors, it returns the identified single node within `input`, or `null` if no node is identified. 
+- If `path` potentially identifies multiple nodes within `input` (by using descendant, wildcard, union, array subset, or filter selectors), it returns an array containing the identified nodes, or an empty array if no node is identified. 
+- If `input` is not a valid JSON value, the function returns `null`.
+- If `path` is `null`, not a valid [JSONPath expression](#Path), or does not match the structure of `input` (for example applying an index selector to a scalar value), the function returns `null`. 
+          </String>
```
HeikoTheissen, line 169:
```suggestion

            The SQL function `JSON_QUERY` returns the stringified version of the extracted JSON value.
          </String>
```

vocabularies/Org.OData.JSON.V1.xml
```diff
@@ -14,64 +14,202 @@
...
+- If `path` only consists of the root identifier followed by name and index selectors and identifies a single scalar JSON value (string, number, boolean, or `null`) within `input`, it returns the identified single value cast to an OData primitive value (see below).
+- If `path` identifies multiple nodes within `input` (by using descendant, wildcard, union, array subset, or filter selectors), identifies an object or array, or does not identify any node, the function returns `null`.
+- If `input` is not a valid JSON value, the function returns `null`.
+- If `path` is `null`, not a valid [JSONPath expression](#Path), or does not match the structure of `input` (for example applying an index selector to a scalar value), the function returns `null`.
+
+If a single non-null scalar JSON value is identified by `path` within `input`, the function returns that value as an OData primitive value of type
+- `Edm.String` if the value is a JSON string
+- `Edm.Boolean` if the value is `true` or `false`
+- `Edm.Decimal` with unspecified precision and floating scale if the value is a JSON number
+          </String>
```
HeikoTheissen, line 198:
```suggestion

            The SQL function `JSON_QUERY` returns a JSON-stringified version of the extracted primitive value.
          </String>
```

HeikoTheissen (commented):
Suggest adding this as implementation hints.

vocabularies/Org.OData.JSON.V1.xml
ralfhandl: I'd rather add this to a new section/document on "implementation options", not to the definition of the OData functions, which may be implemented in a number of different ways on different platforms.

vocabularies/Org.OData.JSON.V1.xml
```diff
@@ -14,64 +14,202 @@
...
+          String="Query a stream value of media type `application/json`, returning an OData primitive value" />
+        <Annotation Term="Core.LongDescription">
+          <String>Extracts a single OData primitive value from the `input` JSON value:
+- If `path` only consists of the root identifier followed by name and index selectors and identifies a single scalar JSON value (string, number, boolean, or `null`) within `input`, it returns the identified single value cast to an OData primitive value (see below).
+- If `path` identifies multiple nodes within `input` (by using descendant, wildcard, union, array subset, or filter selectors), identifies an object or array, or does not identify any node, the function returns `null`.
+- If `input` is not a valid JSON value, the function returns `null`.
+- If `path` is `null`, not a valid [JSONPath expression](#Path), or does not match the structure of `input` (for example applying an index selector to a scalar value), the function returns `null`.
+
+If a single non-null scalar JSON value is identified by `path` within `input`, the function returns that value as an OData primitive value of type
```
HeikoTheissen, line 194:
```suggestion
If null is identified by `path` within `input`, the function returns null.

If a single non-null scalar JSON value is identified by `path` within `input`, the function returns that value as an OData primitive value of type
```

HeikoTheissen (changes requested):
`JSON.value` can return null not only in error cases.

vocabularies/Org.OData.JSON.V1.xml
ralfhandl: This is already part of the first bullet point.

HeikoTheissen: The first bullet point says:
> it returns the identified single value (null) cast to an OData primitive value (see below).

But the list below rules out the case of a null value.

HeikoTheissen (approved):

HeikoTheissen (changes requested):
TC 2024-04-03: Instead of `JSON.value` and `JSON.query`, have only one function with an additional optional parameter, that specifies the type to which the result is cast. If that type is primitive and has an SQL counterpart, the function is converted to the SQL function `JSON_VALUE(..., ... RETURNING ...)`.

mikepizzo:
Alternative: single JSON_QUERY function with optional 3rd parameter that specified return type.  The type of this function would have to be Edm.Untyped; may still have to cast to compose in expression.

Other alternative: have different functions for each type to be returned, thus avoiding the cast.

HeikoTheissen (approved):

sthagen (approved):
LGTM.

vocabularies/Org.OData.JSON.V1.md
```diff
@@ -1,17 +1,190 @@
...
+and the OData JSON format allows [direct embedding of JSON data](https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_StreamProperty) in request and response payloads.
+
+This vocabulary defines a convenience [type for JSON data](#JSON) as well as a term for referencing a [JSON Schema](#Schema) describing the structure of the JSON data.
+
+In addition it defines two functions for [querying](#query) JSON data and using a [primitive value](#value) extracted from JSON data in common expressions, for example in `$filter`, `$orderby`, or `$compute`.
+
+**Example**
+
+The `Employees` entity set has a JSON data property `resume`:
+```json
```
mikepizzo, line 16: Example in the .xml schema should be (or reference) xml.

mikepizzo: TC decided to reference the example rather than copy it into the Vocabulary file.

PR #161, opened 2021-10-08
Temporal - ReturnType: [TimesliceWithPeriod]

HeikoTheissen:
CHANGELOG:
* All temporal actions switch their return type to `Collection(Temporal.TimesliceWithPeriod)`.
* `Temporal.TimesliceWithPeriod/Timeslice` must be a *navigation* property because it has an entity type.
* In the timeline example, `Temporal.ApplicationTimeSupport/Timeline/Period(Start,End)` targets the `history` navigation property in container mode, therefore the path does not include `history/`.
* Upper boundary properties use `max` as their fallback default value.

Following TC agreement on 2021-10-07


ralfhandl (approved):

vocabularies/Org.OData.Temporal.V1.xml
```diff
@@ -149,22 +152,27 @@
...
         </Property>
         <Property Name="PeriodEnd" Type="Edm.PrimitiveType" Nullable="true">
           <Annotation Term="Core.Description" String="Upper boundary of the time slice" />
         </Property>
+        <NavigationProperty Name="Timeslice" Type="Edm.EntityType" Nullable="false">
+          <Annotation Term="Core.AutoExpand" />
+          <Annotation Term="Core.Description" String="A time slice with the same entity type as the binding parameter of the action" />
+          <Annotation Term="Core.LongDescription">
+            <String>The time slice has the same entity set as the binding parameter value.</String>
+          </Annotation>
```
ralfhandl, line 169: I would remove this sentence: the time slices in the request body are "instructions" on how to modify the entity set the action is called on, they aren't full entities.

```suggestion
```

HeikoTheissen: But after `$apply=groupby(...,aggregate(...))`, the result doesn't contain "full" entities either, and still we consider it to have the same `@odata.context` as before. I see it the same way here.

ralfhandl (approved):
Will work until we use TimesliceWithPeriod outside of an action 😄

ralfhandl (approved):

PR #171, opened 2021-11-12
ODATA-1489: add DispositionType to Core.ContentDisposition

ralfhandl:


vocabularies/Org.OData.Core.V1.json
```diff
@@ -512,9 +511,13 @@
         },
         "ContentDispositionType": {
             "$Kind": "ComplexType",
+            "Type": {
+                "$Nullable": true,
```
HeikoTheissen, line 515: What does absence of this property mean? Is there a default value?

vocabularies/Org.OData.Core.V1.json
```diff
@@ -512,9 +511,13 @@
...
+                "@Core.Description": "The disposition type of the binary or stream value, see [RFC 6266, Disposition Type](https://datatracker.ietf.org/doc/html/rfc6266#section-4.2)"
+            },
             "Filename": {
                 "$Nullable": true,
-                "@Core.Description": "The proposed filename for downloading the binary or stream value"
+                "@Core.Description": "The proposed filename for downloading the binary or stream value, see [RFC 6266, Disposition Parameter: 'Filename'](https://datatracker.ietf.org/doc/html/rfc6266#section-4.3)"
```
HeikoTheissen, line 520: `Filename` makes sense only if `Type = "Attachment"`.

ralfhandl: From RFC6266:

>   Depending on the disposition type, this information might be used
>   right away (in the "save as..." interaction caused for the
>   "attachment" disposition type), **or later on (for instance, when the
>   user decides to save the contents of the current page being
>   displayed).**

The "save later" use case combines `Type: inline` with a `Filename`.

vocabularies/Org.OData.Core.V1.json
ralfhandl: I tend towards `DefaultValue="attachment"` because in most cases specifying a `Filename` is done in combination with `attachment`, and would be compatible with the current use of the annotation, only specifying the "download filename".

No default value can also be considered compatible.

vocabularies/Org.OData.Core.V1.xml
```diff
@@ -416,13 +416,16 @@
         <Annotation Term="Core.RequiresType" String="Edm.String" />
       </Term>
 
-      <Term Name="ContentDisposition" Type="Core.ContentDispositionType" AppliesTo="Property Term">
+      <Term Name="ContentDisposition" Type="Core.ContentDispositionType" Nullable="false" AppliesTo="Property Term">
         <Annotation Term="Core.Description" String="The content disposition of a binary or stream property or term" />
       </Term>
 
       <ComplexType Name="ContentDispositionType">
+        <Property Name="Type" Type="Edm.String">
```
HeikoTheissen, line 424:
```suggestion
        <Property Name="Type" Type="Edm.String" DefaultValue="attachment">
```

HeikoTheissen (changes requested):

HeikoTheissen (approved):

PR #176, opened 2021-12-16
ODATA-1459: Undefined grouping properties

HeikoTheissen:
https://issues.oasis-open.org/browse/ODATA-1459


docs/groupKey.js
```diff
@@ -0,0 +1,19 @@
+function groupKey(groupingProperties, instance) {
```
ralfhandl, line 1: Do we have a unit test for this function?

Should we have executable code in our `docs` folder, or rather a [Jupyter notebook](https://github.com/n-riesco/ijavascript) with text and code and results intermingled, literate-programming style, following Donald E. Knuth's example?

HeikoTheissen: The code is not meant to be executed but serves documentary purposes only. And github.com offers nice features like history and edit-by-pull-request, would Jupyter also offer this?

HeikoTheissen: I've nevertheless added a test case, and it immediately spotted superfluous code.

HeikoTheissen:
Agreed to close this with @ralfhandl and @GeraldKrause on 2022-01-14.

PR #183, opened 2022-02-03
Express hierarchy transformations as functions

HeikoTheissen:


vocabularies/Org.OData.Aggregation.V1.xml
```diff
@@ -202,7 +202,7 @@
 
       <ComplexType Name="RecursiveHierarchyType">
         <Property Name="NodeProperty" Type="Edm.PropertyPath" Nullable="false">
-          <Annotation Term="Core.Description" String="Property holding the hierarchy node value" />
+          <Annotation Term="Core.Description" String="Property holding the hierarchy node value. It MUST be of type `Edm.String`." />
```
ralfhandl, line 205: So can we change the type of parameters `Node`, `Ancestor`, `Descendant`, and `Sibling` also to `Edm.String`?

vocabularies/Org.OData.Aggregation.V1.xml
```diff
@@ -266,17 +266,48 @@ given by the `HierarchyNodes` parameter. This specifies a recursive hierarchy th
...
       </Function>
 
+      <Function Name="descendants" IsBound="true">
+        <Annotation Term="Core.Description" String="Returns the node values of descendants in the specified hierarchy of the given nodes with at most the specified distance" />
+        <Parameter Name="Nodes" Type="Collection(Edm.EntityType)" Nullable="false">
+          <Annotation Term="Core.Description" String="The given nodes" />
+        </Parameter>
+        <Parameter Name="NodePropertyPath" Type="Edm.String" Nullable="false">
+          <Annotation Term="Core.Description" String="Property path in the entity type of `Nodes` that addresses the node value" />
+        </Parameter>
```
HeikoTheissen, line 276:
```suggestion
        <Parameter Name="Nodes" Type="Collection(Edm.String)" Nullable="false">
          <Annotation Term="Core.Description" String="Node values of the given nodes" />
        </Parameter>
```
I would much prefer the binding parameter to be a collection of _node values_ rather than a collection of _node entities_, from which the function must pick the node values property by following the `NodePropertyPath`. But, alas, such a collection of strings cannot be computed as a `rootExpr`: `$root/SalesOrganizations/ID` is not a `rootExpr`, because `/ID` is not a `collectionNavigation`.

HeikoTheissen:
Closed after agreement in TC 2022-02-17.

PR #185, opened 2022-03-04
ODATA-1515 - annotate the media entity type to affect its media stream

ralfhandl:


vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -535,13 +535,13 @@ supported:
...
-          <Annotation Term="Core.Description" String="$expand is supported for stream properties and media resources" />
+          <Annotation Term="Core.Description" String="$expand is supported for stream properties and media streams" />
         </Property>
         <Property Name="NonExpandableProperties" Type="Collection(Edm.NavigationPropertyPath)" Nullable="false">
           <Annotation Term="Core.Description" String="These properties cannot be used in expand expressions" />
         </Property>
-        <Property Name="NonExpandableStreamProperties" Type="Collection(Edm.PropertyPath)" Nullable="false">
-          <Annotation Term="Core.Description" String="These stream properties cannot be used in expand expressions" />
+        <Property Name="NonExpandableStreamProperties" Type="Collection(Edm.PropertyPath)" Nullable="true">
+          <Annotation Term="Core.Description" String="These stream properties cannot be used in expand expressions. The special value `null` refers to the media stream of a media entity type." />
```
HeikoTheissen, line 544: This change is not necessary. To disallow `MediaEntity?$expand=$value`, one can annotate
```json
"ns.Container/MediaEntity@Capabilities.ExpandRestrictions": {
  "StreamsExpandable": false
}
```
(unless the media entity type has additional stream properties, which certainly goes against the spirit of media entity types).

HeikoTheissen (changes requested):

vocabularies/Org.OData.Capabilities.V1.xml
ralfhandl: This is exactly the edge case I want to cover 😄

HeikoTheissen (approved):

PR #186, opened 2022-03-18
Digital/Physical -> Food/NonFood

HeikoTheissen:
https://issues.oasis-open.org/browse/ODATA-1468

Adapt example model to new [OData-Aggr, example 63].

### DIFF
--- a/examples/Org.OData.Aggregation.V1.SalesModel-sample.xml
+++ b/examples/Org.OData.Aggregation.V1.SalesModel-sample.xml
@@ -109,6 +109,46 @@
           <NavigationPropertyBinding Path="Time" Target="Time" />
         </EntitySet>
       </EntityContainer>
+      <Annotations Target="SalesModel.Container/Sales">
+        <Annotation Term="ApplySupported">
+          <Record>
+            <PropertyValue Property="Rollup" EnumMember="Aggregation.RollupType/MultipleHierarchies" />
+            <PropertyValue Property="GroupableProperties">
+              <Collection>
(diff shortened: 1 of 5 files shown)


examples/Org.OData.Aggregation.V1.SalesModel-sample.xml
```diff
@@ -109,6 +109,59 @@
...
+                <PropertyPath>SalesOrganization/Name</PropertyPath>
+                <PropertyPath>Product/ID</PropertyPath>
+                <PropertyPath>Product/Name</PropertyPath>
+                <PropertyPath>Product/Color</PropertyPath>
+                <PropertyPath>Product/Category/ID</PropertyPath>
+                <PropertyPath>Product/Category/Name</PropertyPath>
+                <PropertyPath>Customer/ID</PropertyPath>
+                <PropertyPath>Customer/Name</PropertyPath>
+                <PropertyPath>Customer/Country</PropertyPath>
+                <PropertyPath>Time/Date</PropertyPath>
```
HeikoTheissen, line 128: TC 2022-04-14: Examples contain `groupby((Time))`, which is allowed if and only if `groupby((Time/Date))` (the key proeprty) is allowed. `GroupableProperties` only contain the properties, not the navigation properties. Call this out in the vocabulary. Alternatively, allow
```suggestion
                <NavigationPropertyPath>Time</NavigationPropertyPath>
                <PropertyPath>Time/Date</PropertyPath>
```
(changing the vocabulary to 4.01 and `Edm.AnyPropertyPath`).

PR #194, opened 2022-07-29
ErrorResponses: remove accidental nullability

ralfhandl:
A `null` entry in the `ErrorResponses` array would not be useful.


HeikoTheissen:
Why does the change not affect the .json and .md files?

ralfhandl:
> Why does the change not affect the .json and .md files?

For `Type="Collection(...)"` the .json will only contain a `$Nullable` property if the XML source contains the `Nullable` attribute, interpreting an omitted `Nullable` attribute as `false`. Which is correct in most of the cases.

HeikoTheissen (approved):

PR #198, opened 2022-09-12
Fix: warn about undefined aliases

ralfhandl:


HeikoTheissen (approved):
But this does not produce warnings for unknown namespaces in AllowedTerms, right?

ralfhandl:
> But this does not produce warnings for unknown namespaces in AllowedTerms, right?

It already does, now added check in unit test

HeikoTheissen (approved):

PR #191, opened 2022-05-23
Changes resulting from OData-Aggr walkthrough

HeikoTheissen:

PR #204, opened 2022-12-08
Allowed values for transformations and aggregation methods

ralfhandl:


HeikoTheissen (approved):
CS03 transformations must be added in #191.

HeikoTheissen (approved):

PR #214, opened 2023-02-03
Add vocabulary terms to Require Explicit Binding

mikepizzo:
Addresses https://issues.oasis-open.org/browse/ODATA-1538.
-Adds RequiresExplicitBindings vocabulary term definition
-Adds ExplicitOperationBindings vocabulary term definition


vocabularies/Org.OData.Core.V1.xml
```diff
@@ -509,12 +509,19 @@ Any simple identifier | Any type listed in `Validation.OpenPropertyTypeConstrain
         <Annotation Term="Core.LongDescription" String="The annotation value will usually be an expression, e.g. using properties of the binding parameter type for instance-dependent availability, or using properties of a singleton for global availability. The static value `null` means that availability cannot be determined upfront and is instead expressed as an operation advertisement." />
       </Term>
 
-      <TypeDefinition Name="LocalDateTime" UnderlyingType="Edm.String">
+      <Term Name="RequiresExplicitBinding" Type="Core.Tag" DefaultValue="true" AppliesTo="Action, Function">
+		<Annotation Term="Core.Description" String="This bound action or function is only available on model elements annotated with the ExplicitOperationBindings term." />
+      </Term>
+
+      <Term Name="ExplicitOperationBindings" Type="Collection(Edm.String)">
+        <Annotation Term="Core.Description" String="The qualified names of explicitly bound operations that are supported on the target model element. These bindings are in addition to any bindings to the type of the target model element not annotated with RequiresExplicitBinding." />
```
HeikoTheissen, line 517: What are "bindings ... annotated with `RequiresExplicitBinding`"? Should that be "actions/functions annotated with ..."?

mikepizzo: Good catch -- thanks!

vocabularies/Org.OData.Core.V1.xml
```diff
@@ -509,22 +509,23 @@ Any simple identifier | Any type listed in `Validation.OpenPropertyTypeConstrain
...
-		<Annotation Term="Core.Description" String="This bound action or function is only available on model elements annotated with the ExplicitOperationBindings term." />
+      <Term Name="RequiresExplicitBinding" Type="Core.Tag" DefaultValue="true" AppliesTo="Action Function">
+        <Annotation Term="Core.Description" String="This bound action or function is only available on model elements annotated with the ExplicitOperationBindings term." />
       </Term>
 
       <Term Name="ExplicitOperationBindings" Type="Collection(Edm.String)">
-        <Annotation Term="Core.Description" String="The qualified names of explicitly bound operations that are supported on the target model element. These bindings are in addition to any bindings to the type of the target model element not annotated with RequiresExplicitBinding." />
+        <Annotation Term="Core.Description" String="The qualified names of explicitly bound operations that are supported on the target model element. These operations are in addition to any operations bound to the type of the target model element not annotated with RequiresExplicitBinding." />
```
HeikoTheissen, line 517:
```suggestion
        <Annotation Term="Core.Description" String="The qualified names of explicitly bound operations that are supported on the target model element. These operations are in addition to any operations not annotated with RequiresExplicitBinding that are bound to the type of the target model element ." />
```
It's the operation that is "not annotated", not the target model element.

HeikoTheissen (changes requested):

HeikoTheissen (approved):

vocabularies/Org.OData.Core.V1.xml
```diff
@@ -509,15 +509,23 @@ Any simple identifier | Any type listed in `Validation.OpenPropertyTypeConstrain
         <Annotation Term="Core.LongDescription" String="The annotation value will usually be an expression, e.g. using properties of the binding parameter type for instance-dependent availability, or using properties of a singleton for global availability. The static value `null` means that availability cannot be determined upfront and is instead expressed as an operation advertisement." />
       </Term>
 
+      <Term Name="RequiresExplicitBinding" Type="Core.Tag" DefaultValue="true" AppliesTo="Action Function">
+        <Annotation Term="Core.Description" String="This bound action or function is only available on model elements annotated with the ExplicitOperationBindings term." />
+      </Term>
+
+      <Term Name="ExplicitOperationBindings" Type="Collection(Edm.String)">
```
ralfhandl, line 516: I'd introduce a type definition similar to the existing `QualifiedActionName`. 

Also: would this be just the name, allowing all function overloads with a matching binding parameter type, or would you also want to identify specific function overloads?

ralfhandl (changes requested):

vocabularies/Org.OData.Core.V1.xml
```diff
@@ -457,6 +457,10 @@
         <Annotation Term="Core.Description" String="The qualified name of an action in scope." />
       </TypeDefinition>
 
+      <TypeDefinition Name="QualifiedOperationName" UnderlyingType="Edm.String">
+        <Annotation Term="Core.Description" String="The qualified name of an action or function in scope. For a Function, the qualified function name may optionally be followed by parentheses containing the comma-separated list of parameter types, in the order of their definition, to identify a single function overload." />
```
HeikoTheissen, line 461:
```suggestion
        <Annotation Term="Core.Description" String="The qualified name of an action or function in scope, optionally followed by parentheses containing the comma-separated list of parameter types, in the order of their definition, to identify a single function overload." />
```
Otherwise it sounds to me as if the parenthesised expression _followed_ the `QualifiedOperationName` rather than being part of it.

HeikoTheissen (changes requested):

vocabularies/Org.OData.Core.V1.xml
mikepizzo: Bound actions only support overloading by binding type, so for this usage there's no value in appending the parameters for the action case; only the function case.

So how about: "The qualified name of a bound action or function in scope. The qualified function name may include trailing parentheses containing the comma-separated list of parameter types, in the order of their definition, to identify a single function overload."

vocabularies/Org.OData.Core.V1.xml
```diff
@@ -457,6 +457,10 @@
         <Annotation Term="Core.Description" String="The qualified name of an action in scope." />
       </TypeDefinition>
 
+      <TypeDefinition Name="QualifiedBoundOperationName" UnderlyingType="Edm.String">
+        <Annotation Term="Core.Description" String="The qualified name of a bound action or function in scope. The qualified function name may include trailing parentheses containing the comma-separated list of parameter types, in the order of their definition, to identify a single function overload." />
```
ralfhandl, line 461: While the current use of this type definition does not need to identify action overloads, a generic type definition may want to allow it.

vocabularies/Org.OData.Core.V1.xml
HeikoTheissen: I did not mean to allow overloading of actions, I thought this would be clear from:
> optionally followed by parentheses containing the comma-separated list of parameter types, in the order of their definition, to identify a single _function_ overload.

But your suggestion is clearer than mine.

vocabularies/Org.OData.Core.V1.xml
mikepizzo: In order to support multiple action overloads, the type definition would need the ability to specify the binding type. However, in our intended usage, the binding type is determined by the type of the element being annotated.

We discussed in our TC meeting and felt it was simpler to restrict the type definition to be used in cases where the binding parameter type is already determined (thus, the QualifiedBoundOperationName). If, in the future, we need to be able to specify a binding parameter type to identify other action overloads (or functions bound to other types), we can add a new typedef.

We added a new long description to describe the 3 cases:
-action name to specify the single action bound to the specified type
-function name to specify all function overloads bound to the specified type
-function name followed by parameters, to indicate a single function overload bound to the specified type.

HeikoTheissen (changes requested):

vocabularies/Org.OData.Core.V1.xml
```diff
@@ -457,6 +457,11 @@
         <Annotation Term="Core.Description" String="The qualified name of an action in scope." />
       </TypeDefinition>
 
+      <TypeDefinition Name="QualifiedBoundOperationName" UnderlyingType="Edm.String">
+        <Annotation Term="Core.Description" String="The qualified name of a bound action or function in scope." />
```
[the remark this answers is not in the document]
mikepizzo: Seems we're pretty inconsistent. I modelled these after the existing "QualifiedActionName" and "QualifiedTypeName", which do have a full stop after their respective descriptions.

Should we do a PR to make these consistent across all of the definitions in this (and other) vocabularies?

vocabularies/Org.OData.Core.V1.xml
```diff
@@ -457,6 +457,11 @@
         <Annotation Term="Core.Description" String="The qualified name of an action in scope." />
       </TypeDefinition>
 
+      <TypeDefinition Name="QualifiedBoundOperationName" UnderlyingType="Edm.String">
+        <Annotation Term="Core.Description" String="The qualified name of a bound action or function in scope." />
+        <Annotation Term="Core.LongDescription" String="Either the qualified name of an action, to indicate the single bound overload with the specified binding parameter type, the qualified name of a function, to indicate all bound overloads with the specified binding parameter type, or the qualified name of a function followed by parentheses containing a comma-separated list of parameter types, in the order of their definition, to identify a single function overload with the first (binding) parameter matching the specified parameter type." />
```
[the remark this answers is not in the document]
mikepizzo: I debated this, but I don't see anywhere that we mention the use of HTML tags in long descriptions, nor did I see other examples of such usage.

HeikoTheissen (approved):

vocabularies/Org.OData.Core.V1.xml
ralfhandl, line 462:
```suggestion
        <Annotation Term="Core.LongDescription">
          <String>Either
- the qualified name of an action, to indicate the single bound overload with the specified binding parameter type, 
- the qualified name of a function, to indicate all bound overloads with the specified binding parameter type, or 
- the qualified name of a function followed by parentheses containing a comma-separated list of parameter types, in the order of their definition, to identify a single function overload with the first (binding) parameter matching the specified parameter type.
        </String>
        </Annotation>
```

So far we rely on XML retaining line-breaks within element text, combined with Markdown treatment of line-breaks.

ralfhandl (commented):
Suggested formatting improvement

PR #219, opened 2023-04-26
Don't repeat redefined properties in derived type

HeikoTheissen:
When a property in a base type is redefined in a derived type (by specializing the property type from `Edm.Untyped` to `Edm.PrimitiveType`, say), both definitions are shown in the Markdown:
https://github.com/SAP/odata-vocabularies/blob/c5fd15aa19a3f8413ace9a7186eccd054ef60a7e/vocabularies/UI.md?plain=1#L975-L976

- [x] add unit test to keep this feature alive
- [x] bump package version to 0.5
- [x] bump dependencies

### DIFF
--- a/lib/csdl2markdown.js
+++ b/lib/csdl2markdown.js
@@ -428,7 +428,7 @@ module.exports.csdl2markdown = function (filename, csdl, referenced = {}) {
 
     if (!type) return lines;
 
-    var names = Object.keys(type).filter((name) => isIdentifier(name));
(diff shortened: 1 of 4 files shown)


ralfhandl (changes requested):
The effects of the code change can only be seen in the SAP/odata-vocabularies repo, no unit test here to prevent accidentally breaking this feature

ralfhandl (approved):

ralfhandl (approved):

PR #228, opened 2023-09-22
Functions and actions as interfaces

HeikoTheissen:
[ODATA-1600](https://issues.oasis-open.org/browse/ODATA-1600)


vocabularies/Org.OData.Core.V1.xml
```diff
@@ -558,6 +558,18 @@ Any simple identifier | Any type listed in `Validation.OpenPropertyTypeConstrain
...
 
+      <Term Name="IsInterface" Type="Core.Tag" Nullable="false" DefaultValue="true" AppliesTo="Action Function">
+        <Annotation Term="Core.Description" String="The tagged action or function cannot be invoked, it serves only as an interface" />
+        <Annotation Term="Core.LongDescription">
+          <String>Services that want to implement the interface define an action or function with the same signature
+          or a specialized signature that replaces abstract types like `Edm.EntityType` with concrete types.
+          These function MUST be annotated with [`ImplementsInterface`](#ImplementsInterface).</String>
+        </Annotation>
+      </Term>
+      <Term Name="ImplementsInterface" Type="Edm.String" Nullable="false" AppliesTo="Action Function">
```
ralfhandl, line 569: How about using a more specific type such as [QualifiedBoundOperationName](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.xml#L460:~:text=%3CTypeDefinition%20Name=%22-,QualifiedBoundOperationName,-%22) or a similar new one without the "Bound" infix?

HeikoTheissen: A type for exactly one usage?

examples/Org.OData.Core.V1.IsSignature-equivalence.md
HeikoTheissen, line 1: Is this a valid example for allowed differences between signature and implementation?

lib/transform.js
```diff
@@ -72,14 +72,24 @@ fs.readdirSync(exampleFolder)
       const json = csdl.xml2json(xml, { strict: true });
       fs.writeFileSync(
         exampleFolder + example + ".json",
-        JSON.stringify(json, null, 4)
+        JSON.stringify(json, null, 4),
       );
+      if (!example.endsWith("-sample")) {
```
ralfhandl, line 77: Why not use a suffix such as `-vocabsample` to trigger vocab-style markdown conversion? The vocab-style markdown conversion is the new exception, so make it an exception and not the new default.

ralfhandl (changes requested):

ralfhandl (approved):

ralfhandl (approved):

PR #232, opened 2024-01-22
ODATA-1631 - Validation.Constraint also on NavigationProperty

ralfhandl:


vocabularies/Org.OData.Validation.V1.xml
```diff
@@ -109,7 +109,7 @@
         <Annotation Term="Core.Description" String="The value of the annotated property, parameter, or term must be an integer multiple of this positive value. For temporal types, the value is measured in seconds." />
       </Term>
 
-      <Term Name="Constraint" Type="Validation.ConstraintType" Nullable="false" AppliesTo="Property EntityType ComplexType">
+      <Term Name="Constraint" Type="Validation.ConstraintType" Nullable="false" AppliesTo="Property NavigationProperty EntityType ComplexType">
```
HeikoTheissen, line 112: Can this also apply to a Term?

HeikoTheissen: Sorry, I mixed this up with other capabilities terms, like `Minimum`.

HeikoTheissen (approved):

vocabularies/Org.OData.Validation.V1.xml
ralfhandl: I'm not opposed to adding `Parameter` and `Term` just to be consistent with the specialized constraints: should we do that now, or wait for a concrete use case?

HeikoTheissen: `Constraint` is about error messages presented to the user. This makes sense for `Parameter` (which the user may have entered on the UI), but not for `Term`.

ralfhandl: Added `Parameter`

mikepizzo (approved):

PR #238, opened 2024-02-22
ODATA-1643: Discourage use of implicit grant

ralfhandl:
See https://github.com/OAI/OpenAPI-Specification/pull/3584

### DIFF
--- a/vocabularies/Org.OData.Authorization.V1.xml
+++ b/vocabularies/Org.OData.Authorization.V1.xml
@@ -123,6 +123,7 @@
       </ComplexType>
 
       <ComplexType Name="OAuth2Implicit" BaseType="Authorization.OAuthAuthorization">
+        <Annotation Term="Core.Description" String="Security note: OAuth2 implicit grant is considered to be not secure and should not be used by clients, see [OAuth 2.0 Security Best Current Practice](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics.html#name-implicit-grant)." />
         <Property Name="AuthorizationUrl" Type="Edm.String" Nullable="false">
           <Annotation Term="Core.Description" String="Authorization URL" />
           <Annotation Term="Core.IsURL" />
--- a/vocabularies/Org.OData.Authorization.V1.json
+++ b/vocabularies/Org.OData.Authorization.V1.json
@@ -113,6 +113,7 @@
         "OAuth2Implicit": {
             "$Kind": "ComplexType",
             "$BaseType": "Authorization.OAuthAuthorization",
+            "@Core.Description": "Security note: OAuth2 implicit grant is considered to be not secure and should not be used by clients, see [OAuth 2.0 Security Best Current Practice](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics.html#name-implicit-grant).",
             "AuthorizationUrl": {
                 "@Core.Description": "Authorization URL",
                 "@Core.IsURL": true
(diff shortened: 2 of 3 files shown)


vocabularies/Org.OData.Authorization.V1.xml
```diff
@@ -123,6 +123,7 @@
       </ComplexType>
 
       <ComplexType Name="OAuth2Implicit" BaseType="Authorization.OAuthAuthorization">
+        <Annotation Term="Core.Description" String="Security note: OAuth2 implicit grant is considered to be not secure and should not be used by clients, see [OAuth 2.0 Security Best Current Practice](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics.html#name-implicit-grant)." />
```
HeikoTheissen, line 126:
```suggestion
        <Annotation Term="Core.Description" String="Security note: OAuth2 implicit grant should only be used by clients in combination with [PKCE](https://www.rfc-editor.org/rfc/rfc7636), see [OAuth 2.0 Security Best Current Practice](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics.html#name-implicit-grant)." />
```
I think that implicit grant with PKCE is acceptable.

HeikoTheissen (changes requested):

vocabularies/Org.OData.Authorization.V1.xml
ralfhandl: https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/05-Authorization_Testing/05-Testing_for_OAuth_Weaknesses states that

> Two flows will be deprecated in the release of [OAuth2.1](https://oauth.net/2.1/), and their usage is not recommended:
> * [Implicit Flow*](https://oauth.net/2/grant-types/implicit/): PKCE’s secure implementation renders this flow obsolete. Prior to PKCE, the implicit flow was used by client-side applications such as [single page applications](https://en.wikipedia.org/wiki/Single-page_application) since [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) relaxed the [same-origin policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy) for sites to inter-communicate. For more information on why the implicit grant is not recommended, review this [section](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-2.1.2).

HeikoTheissen: Is the PKCE flow described in https://www.rfc-editor.org/rfc/rfc7636 recommended? Is that the recommended successor of the implicit flow?

ralfhandl: Good questions, let's ask a security expert.

HeikoTheissen: The experts recommend PKCE, does this mean that we encourage [browser-based OAuth 2.0 clients](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-browser-based-apps#name-browser-based-oauth-20-clie)?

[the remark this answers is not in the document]
ralfhandl: I'm hesitant to add a recommendation specifically for browser-based clients. The referenced section on implicit grant already recommends alternatives for all kinds of clients

> Clients SHOULD instead use the response type "code" (aka authorization code grant type) as specified in [Section 2.1.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics.html#ac) or any other response type that causes the authorization server to issue access tokens in the token response, such as the "code id_token" response type.

I'd just leave the text as is.

HeikoTheissen (approved):

PR #241, opened 2024-04-04
Alternative search syntax

HeikoTheissen:


vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -570,11 +570,15 @@ supported:
...
         <Property Name="Searchable" Type="Edm.Boolean" Nullable="false" DefaultValue="true">
-          <Annotation Term="Core.Description" String="$search is supported" />
+          <Annotation Term="Core.Description" String="$search is supported with the standard syntax [OData-URL, section 5.1.8.1](https://docs.oasis-open.org/odata/odata/v4.02/odata-v4.02-part2-url-conventions.html#SearchExpressions)" />
         </Property>
         <Property Name="UnsupportedExpressions" Type="Capabilities.SearchExpressions" Nullable="false" DefaultValue="none">
           <Annotation Term="Core.Description" String="Expressions not supported in $search" />
         </Property>
+        <Property Name="AlternativeSearchSyntax" Type="Edm.String" Nullable="true">
+          <Annotation Term="Core.IsURL" />
+          <Annotation Term="Core.Description" String="URI of the alternative $search syntax supported by the service" />
```
ralfhandl, line 580:
```suggestion
          <Annotation Term="Core.Description" String="URL of the alternative $search syntax supported by the service" />
```

I'd prefer requiring a resolvable Locator here instead of a mere Identifier.

ralfhandl (commented):
Minor change proposal

ralfhandl (approved):

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -570,11 +570,15 @@ supported:
...
         <Property Name="UnsupportedExpressions" Type="Capabilities.SearchExpressions" Nullable="false" DefaultValue="none">
           <Annotation Term="Core.Description" String="Expressions not supported in $search" />
         </Property>
+        <Property Name="AlternativeSearchSyntax" Type="Edm.String" Nullable="true">
+          <Annotation Term="Core.IsURL" />
+          <Annotation Term="Core.Description" String="URL of the alternative $search syntax supported by the service" />
+        </Property>
       </ComplexType>
       <EnumType Name="SearchExpressions" IsFlags="true">
         <Member Name="none" Value="0">
```
ralfhandl, line 584: Given that the only use of this enum type is for _unsupported_ expressions, the meaning of `none` should be "no restrictions"

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -570,15 +570,19 @@ supported:
...
         </Property>
         <Property Name="UnsupportedExpressions" Type="Capabilities.SearchExpressions" Nullable="false" DefaultValue="none">
           <Annotation Term="Core.Description" String="Expressions not supported in $search" />
         </Property>
+        <Property Name="SearchSyntax" Type="Edm.String" Nullable="true">
+          <Annotation Term="Core.IsURL" />
+          <Annotation Term="Core.Description" String="URL of the $search syntax supported by the service if this deviates from the standard syntax [OData-URL, section 5.1.8.1](https://docs.oasis-open.org/odata/odata/v4.02/odata-v4.02-part2-url-conventions.html#SearchExpressions)" />
+        </Property>
       </ComplexType>
       <EnumType Name="SearchExpressions" IsFlags="true">
```
ralfhandl, line 583: Can we add 
~~~xml
        <Member Name="all" Value="31">
          <Annotation Term="Core.Description" String="Only plain search terms are supported" />
        </Member>
~~~

HeikoTheissen: We can, but that does not mean that NOT may be interpreted as a term to be matched. See https://github.com/oasis-tcs/odata-specs/pull/1918/files#r1570163982

PR #252, opened 2024-04-11
Bulleted lists in tables

HeikoTheissen:


ralfhandl:
Actually I thought we could to this automatically in function https://github.com/oasis-tcs/odata-vocabularies/blob/1887311d443c40067a3285f3e00ef077ae21c117/lib/csdl2markdown.js#L661-L668

HeikoTheissen:
> Actually I thought we could to this automatically in function
> 
> https://github.com/oasis-tcs/odata-vocabularies/blob/1887311d443c40067a3285f3e00ef077ae21c117/lib/csdl2markdown.js#L661-L668

That's like reimplementing Pandoc. I started down that road, but gave it up.

ralfhandl:
> let's ask the author to do it _once_

So the author has to know which `LongDescription` we render into a paragraph and which into a table cell 🤔?

Maybe the problem actually is using tables instead of sub(sub)sections.

HeikoTheissen:
> Actually I thought we could to this automatically in function
> 
> https://github.com/oasis-tcs/odata-vocabularies/blob/1887311d443c40067a3285f3e00ef077ae21c117/lib/csdl2markdown.js#L661-L668

Pandoc on the fly works!

ralfhandl:
Feel free to merge as this PR does not change the vocabulary text

PR #254, opened 2024-05-08
Clarification regarding Validation.Pattern

HeikoTheissen:
The annotated property must match the regular expression _in its entirety_, because a regular expression like `\d+` (one or more digits) is also matched by the string `7up`, just not in its entirety.

### DIFF
--- a/vocabularies/Org.OData.Validation.V1.json
+++ b/vocabularies/Org.OData.Validation.V1.json
@@ -34,7 +34,8 @@
                 "Parameter",
                 "Term"
             ],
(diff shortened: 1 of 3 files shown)


ralfhandl (changes requested):
The current behavior is intended, Validation.Pattern is the counterpart to the JSON Schema keyword [`pattern`](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-6.3.3), which explicitly states that
> regular expressions are not implicitly anchored

If someone wants a prefix, suffix, or full match, they can anchor the regular expression using `^` and `$`, for example `^\d+$` will only match if the string consists entirely of digits and has a minimum length of 1.

PR #255, opened 2024-05-15
Generate Vocabularies document

HeikoTheissen:


ralfhandl:
Why not put all spec documents into the `odata-specs` repo?

PR #257, opened 2024-06-14
Text fragments do not work reliably

HeikoTheissen:
The text fragments we introduced starting with #139 do not work in the XML documents served by github.io. They are served without `Content-Type`, but Google Chrome now displays them as XML and ignores text fragments.

This PR removes the text fragments from the source links in our Markdown documentation. On github.com the `#L123` fragments work. During the github.io build, the relative source links are replaced with absolute links to github.io.


ralfhandl (approved):
Looks good

PR #262, opened 2024-07-05
Update template.html - remove polyfill

ralfhandl:
See
* https://sansec.io/research/polyfill-supply-chain-attack
* https://www.youtube.com/watch?v=ILvNG1STUZU


HeikoTheissen:
@ralfhandl: Pandoc has replaced this polyfill.io URL with https://cdnjs.cloudflare.com/polyfill/v3/polyfill.min.js?features=es6. I suggest we do the same here.

PR #284, opened 2024-11-25
Support services to decide responding asynchronously

GeraldKrause:
Fixes https://github.com/oasis-tcs/odata-specs/issues/2015

### Changes

- Added new term `AsynchronousRequestsRequired`
  - To support annotating this term to only a subset of all resources of the service, `AppliesTo` mentions all relevant resources
- The fine-grained `AppliesTo` leads to an extension of the `AppliesTo` in the base term `AsynchronousRequestsSupported`

### DIFF
--- a/vocabularies/Org.OData.Capabilities.V1.json
+++ b/vocabularies/Org.OData.Capabilities.V1.json
@@ -92,9 +92,29 @@
             "$Type": "Core.Tag",
             "$DefaultValue": true,
             "$AppliesTo": [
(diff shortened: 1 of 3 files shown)


HeikoTheissen (changes requested):
Copy error

HeikoTheissen (approved):

mikepizzo:
Decided to move to V5 where we could introduce this in a cleaner fashion.

PR #285, opened 2024-11-27
Severity in Validation.Constraint

HeikoTheissen:


vocabularies/Org.OData.Validation.V1.xml
```diff
@@ -117,6 +117,9 @@
           <Annotation Term="Core.IsLanguageDependent" />
           <Annotation Term="Core.Description" String="Human-readable message that can be shown to end users if the constraint is not fulfilled" />
         </Property>
+        <Property Name="MessageSeverity" Type="Core.MessageSeverity" Nullable="true">
```
ralfhandl, line 120: In Core.MessageType we only call it `severity`, so `Severity` here to match case would also be fine. Either way.

ralfhandl:
Constraints currently also express what clients can expect to receive, and we don't want to break that.

Better add a new term, and probably in an SAP vocabulary.

PR #286, opened 2025-01-09
All `NavigationPropertyRestriction`s nullable

HeikoTheissen:
Fixes https://github.com/oasis-tcs/odata-openapi/issues/322

### DIFF
--- a/vocabularies/Org.OData.Capabilities.V1.xml
+++ b/vocabularies/Org.OData.Capabilities.V1.xml
@@ -272,16 +272,16 @@ supported:
         <Property Name="SortRestrictions" Type="Capabilities.SortRestrictionsType">
           <Annotation Term="Core.Description" String="Restrictions on orderby expressions" />
         </Property>
-        <Property Name="TopSupported" Type="Edm.Boolean">
+        <Property Name="TopSupported" Type="Edm.Boolean" Nullable="true">
           <Annotation Term="Core.Description" String="Supports $top" />
         </Property>
-        <Property Name="SkipSupported" Type="Edm.Boolean">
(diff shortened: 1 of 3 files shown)


vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -272,16 +272,16 @@ supported:
         <Property Name="SortRestrictions" Type="Capabilities.SortRestrictionsType">
           <Annotation Term="Core.Description" String="Restrictions on orderby expressions" />
         </Property>
-        <Property Name="TopSupported" Type="Edm.Boolean" Nullable="false" DefaultValue="true">
+        <Property Name="TopSupported" Type="Edm.Boolean">
```
ralfhandl, line 275:
```suggestion
        <Property Name="TopSupported" Type="Edm.Boolean" Nullable="true">
```
Better state explicitly that we mean it and didn't just forget it.

And maybe describe what null/omission means.

HeikoTheissen: > Better state explicitly that we mean it and didn't just forget it.

Only for these two, or also for all others?

> And maybe describe what null/omission means.

Isn't that already said?
https://github.com/oasis-tcs/odata-vocabularies/blob/e1a12884052c5b2fe9d535dc1a3918e4d134a5d9/vocabularies/Org.OData.Capabilities.V1.xml#L233

ralfhandl: All four that were touched in this PR, please

ralfhandl (approved):

HeikoTheissen:
TC 2025-01-22 decided against this incompatible change.

PR #290, opened 2025-02-20
Core.OptionalParameter/DefaultsToEmptyCollection

HeikoTheissen:
Fixes oasis-tcs/odata-specs#2045

### DIFF
--- a/vocabularies/Org.OData.Core.V1.xml
+++ b/vocabularies/Org.OData.Core.V1.xml
@@ -513,9 +513,9 @@ Any simple identifier | Any type listed in `Validation.OpenPropertyTypeConstrain
           <Annotation Term="Core.Description" String="Default value for an optional parameter of primitive type, using the same rules as the `cast` function in URLs" />
           <Annotation Term="Core.LongDescription" String="If no explicit DefaultValue is specified, the service is free on how to interpret omitting the parameter from the request. For example, a service might interpret an omitted optional parameter `KeyDate` as having the current date." />
         </Property>
-        <Property Name="NonPrimitiveDefaultValue" Type="Edm.Untyped" Nullable="true">
-          <Annotation Term="Core.Description" String="Default value for an optional parameter of non-primitive type" />
-          <Annotation Term="Core.LongDescription" String="If no explicit NonPrimitiveDefaultValue is specified, the service is free on how to interpret omitting the parameter from the request." />
+        <Property Name="DefaultsToEmptyCollection" Type="Core.Tag" Nullable="false" DefaultValue="false">
+          <Annotation Term="Core.Description" String="Whether an optional collection-valued parameter defaults to an empty collection" />
+          <Annotation Term="Core.LongDescription" String="If this is false, the service is free on how to interpret omitting the parameter from the request." />
         </Property>
       </ComplexType>
 
--- a/vocabularies/Org.OData.Core.V1.xml
+++ b/vocabularies/Org.OData.Core.V1.xml
@@ -510,11 +510,11 @@ Any simple identifier | Any type listed in `Validation.OpenPropertyTypeConstrain
       </Term>
       <ComplexType Name="OptionalParameterType">
         <Property Name="DefaultValue" Type="Edm.String" Nullable="true">
-          <Annotation Term="Core.Description" String="Default value for an optional parameter of primitive or enumeration type, using the same rules as the `cast` function in URLs" />
+          <Annotation Term="Core.Description" String="Default value for an optional parameter of primitive type, using the same rules as the `cast` function in URLs" />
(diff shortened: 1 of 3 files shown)


vocabularies/Org.OData.Core.V1.xml
```diff
@@ -510,9 +510,13 @@ Any simple identifier | Any type listed in `Validation.OpenPropertyTypeConstrain
       </Term>
       <ComplexType Name="OptionalParameterType">
         <Property Name="DefaultValue" Type="Edm.String" Nullable="true">
-          <Annotation Term="Core.Description" String="Default value for an optional parameter of primitive or enumeration type, using the same rules as the `cast` function in URLs." />
+          <Annotation Term="Core.Description" String="Default value for an optional parameter of primitive or enumeration type, using the same rules as the `cast` function in URLs" />
```
ralfhandl, line 513:
```suggestion
          <Annotation Term="Core.Description" String="Default value for an optional parameter of primitive type, using the same rules as the `cast` function in URLs" />
```
[OData Protocol, chapter 3 Data Model](https://docs.oasis-open.org/odata/odata/v4.02/csd01/part1-protocol/odata-v4.02-csd01-part1-protocol.html#DataModel) states that
> Enumeration types are named primitive types whose values are named constants with underlying integer values.

vocabularies/Org.OData.Core.V1.xml
```diff
@@ -510,9 +510,13 @@ Any simple identifier | Any type listed in `Validation.OpenPropertyTypeConstrain
       </Term>
       <ComplexType Name="OptionalParameterType">
         <Property Name="DefaultValue" Type="Edm.String" Nullable="true">
-          <Annotation Term="Core.Description" String="Default value for an optional parameter of primitive or enumeration type, using the same rules as the `cast` function in URLs." />
+          <Annotation Term="Core.Description" String="Default value for an optional parameter of primitive or enumeration type, using the same rules as the `cast` function in URLs" />
           <Annotation Term="Core.LongDescription" String="If no explicit DefaultValue is specified, the service is free on how to interpret omitting the parameter from the request. For example, a service might interpret an omitted optional parameter `KeyDate` as having the current date." />
         </Property>
+        <Property Name="NonPrimitiveDefaultValue" Type="Edm.Untyped" Nullable="true">
+          <Annotation Term="Core.Description" String="Default value for an optional parameter of non-primitive non-enumeration type" />
```
ralfhandl, line 517:
```suggestion
          <Annotation Term="Core.Description" String="Default value for an optional parameter of non-primitive type" />

```

ralfhandl (changes requested):
Simplify text

ralfhandl (approved):

vocabularies/Org.OData.Core.V1.xml
```diff
@@ -510,9 +510,13 @@ Any simple identifier | Any type listed in `Validation.OpenPropertyTypeConstrain
       </Term>
       <ComplexType Name="OptionalParameterType">
         <Property Name="DefaultValue" Type="Edm.String" Nullable="true">
-          <Annotation Term="Core.Description" String="Default value for an optional parameter of primitive or enumeration type, using the same rules as the `cast` function in URLs." />
+          <Annotation Term="Core.Description" String="Default value for an optional parameter of primitive type, using the same rules as the `cast` function in URLs" />
           <Annotation Term="Core.LongDescription" String="If no explicit DefaultValue is specified, the service is free on how to interpret omitting the parameter from the request. For example, a service might interpret an omitted optional parameter `KeyDate` as having the current date." />
         </Property>
+        <Property Name="NonPrimitiveDefaultValue" Type="Edm.Untyped" Nullable="true">
```
ralfhandl, line 516: The vocabulary currently uses OData 4.0, which doesn't have Edm.Untyped:
https://github.com/oasis-tcs/odata-vocabularies/blob/59e19bb1eaf27e88037aacf9a04a2b8562c9997e/vocabularies/Org.OData.Core.V1.xml#L41

Do we have to create a new vocabulary `Org.OData.Core.V2`?

ralfhandl (changes requested):

vocabularies/Org.OData.Core.V1.xml
HeikoTheissen: Or use `Type=Collection(Edm.String)` which allows at least specifiying an empty collection as default.

HeikoTheissen: > Do we have to create a new vocabulary `Org.OData.Core.V2`?

That's too heavy a tool for this small problem.

ralfhandl (approved):

HeikoTheissen:
Not needed, as discussed in TC 2025-02-26.

PR #294, opened 2025-06-03
Avoid flickering when navigating from dark mode to dark mode

HeikoTheissen:
When navigating between pages on https://oasis-tcs.github.io/odata-vocabularies in dark mode, the target page is briefly rendered in light mode and then switched to dark mode, causing unwanted flickering.


ralfhandl:
How to test this?

If you have tested it and it works, please just merge it.

HeikoTheissen:
Tested with `npm run pages` followed by `npm run serve-pages`.

PR #295, opened 2025-06-04
$expand=@Core.DoNotExpand

### LINKED ISSUE #2031: Clarify extent of return response from a deep update
#363 clarified, for a deep insert, if a response is returned then it must be expanded to include at least the entities and properties specified in the request.

We make no such statement for deep update, which can include created nested entities.

HeikoTheissen:
Fixes oasis-tcs/odata-specs#2031


HeikoTheissen:
Does the TC still want to pursue this? SAP needs something like this in use cases that are not covered by oasis-tcs/odata-specs#2223.

mikepizzo:
Both Microsoft and IBM would break if an (unknown) annotation is specified in $select or $expand.  They can fix their implementations, but this does suggest that our trick is not as backward compatible as we had hoped.

If $expanding a special annotation doesn't preserve backward compatibility, we could explore alternate suggestion that might be more elegant, such as:
 * an empty expand list: $expand=
 * just $expand <= a little wierd
 * $expand=false   <= may conflict with a navigation property name
 * $noExpand <= user could specify both $expand and $noExpand
 * $expand=-
 * $expand=null  <=technically could be a property name
   
Would we do the same for $select?  Hubert would like if we did.

General preference for empty list for both $select and $expand.

mikepizzo:
See updated resolution to https://github.com/oasis-tcs/odata-specs/issues/2031

mikepizzo (changes requested):
After discussion in TC, we agreed that the proper way to represent this would be through an empty select or expand list.  See resolution to original issue.

PR #304, opened 2025-09-25
Disallow conflicting restrictions for the same capability

### LINKED ISSUE #298: Relative precedence of "restrictions specific to a path"
https://github.com/oasis-tcs/odata-vocabularies/blob/77a6fb3348444742a712dc499d6affec2aa0ad59/vocabularies/Org.OData.Capabilities.V1.xml#L233

What does the phrase "restrictions specific to a path" mean? Paths can occur
- in an external annotation target of the form `ns.container/EntitySet/NavProp1/.../NavPropN`
- in `Capabilities.NavigationRestrictions/RestrictedProperties/NavigationProperty`
- in a mixture of both.

What is the relative precedence of restriction annotations that use these ways to specify a path?

HeikoTheissen:
Fixes #298


vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -108,13 +108,22 @@ supported:
 - Updatability
 - Deletability
 
-Capabilities annotations whose external target path is a canonical collection (in the
+Capabilities annotations targeting a canonical collection (in the
 sense of [OData-Protocol, section 10](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#ContextURL))
 apply to every entity in that collection, even if it is addressed via a non-canonical path.
```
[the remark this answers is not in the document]
mikepizzo: NavigationRestrictions that follow a canonical path from an entity set or canonical path define defaults for all references to those instances unless more specific path annotations are defined.

vocabularies/Org.OData.Capabilities.V1.md
```diff
@@ -53,13 +53,13 @@ Capabilities annotations can be given through three mechanisms:
...
-even when it is addressed via a different, non-canonical, path.
+If T contains no non-containment navigation property, the capabilities annotation
+applies to every entity that can be addressed by a resource path matching T,
+even when it is addressed via another resource path.
 
-Non-canonical capabilities annotations (where T contains at least one non-containment navigation property)
+Capabilities annotations where T contains at least one non-containment navigation property
 are specific to resource paths matching T; their restrictions take precedence over restrictions
-from canonical capabilities annotations.
+from capabilities annotations where T contains no non-containment navigation property.
```
HeikoTheissen: Bring back "canonical", but only as an attribute of "resource path".

vocabularies/Org.OData.Capabilities.V1.xml
```diff
@@ -120,13 +120,13 @@ Capabilities annotations can be given through three mechanisms:
 It is an error to specify conflicting restrictions of the same capability with the same T.
 A resource path _matches_ T if it contains the same non-container, non-key segments in the same order as T.
 
-If T contains no non-containment navigation property, the capabilities annotation is called _canonical_;
-it then applies to every entity that can be addressed by a resource path matching T,
-even when it is addressed via a different, non-canonical, path.
+If T contains no non-containment navigation property, the capabilities annotation
+applies to every entity that can be addressed by a resource path matching T,
+even when it is addressed via another resource path.
```
mikepizzo:
```suggestion
If T matches a canonical resource path, the capabilities annotation applies to every entity that can be addressed by a resource path matching T, even when it is addressed via another resource path.
```

PR #320, opened 2026-04-01
Use of Core.RequiresExplicitBinding for dynamic operations

### LINKED ISSUE #270: Should we support dynamic functions/actions?
Currently we don't say whether or not we support dynamic actions/functions.

Currently, clients rely on metadata to describe parameters that can be passed to a function/action, so we would either have to augment the function/action element to describe parameters (and return type?) or restrict dynamic actions/functions to parameterless versions.

Imported from [ODATA-579](https://issues.oasis-open.org/browse/ODATA-579)

mikepizzo:
Update definition of Core.RequiresExplicitBinding to include instances explicitly advertised in payload.

Fixes https://github.com/oasis-tcs/odata-specs/issues/270


mikepizzo:
Heiko will fix so that this contains only the change relevant to this issue (the first commit)

PR #322, opened 2026-04-11
Add JsonStreamingSupport term

mikepizzo:
Add new term for JSON Streaming Support.
Note: considered adding a single JsonSupport term for all various json-specific format support features, but decided to keep separate for streaming.


mikepizzo:
We decided not to add this term at this time as requiring existing clients to comply with ordering requirements for json request payloads would be a breaking change.

PR #333, opened 2026-08-05
Clarify Minor Version handling of ODataVersion vocabulary term

### LINKED ISSUE #2231: How is ODataVersion vocabulary term intended to be used/confirmed?
How is ODataVersion annotation intended to be used? Currently the text says:

`
A space-separated list of supported versions of the OData Protocol. Note that 4.0 is implied by 4.01 and does not need to be separately listed.
`

So, what is the direction to the client? If they currently check for 4.01 (and imply that to mean 4.0 as well), and we start returning 4.02 instead, won't that break them? Or should clients check for major version =4 and minor version >= {latest supported version} and trust that we will continue to make minor version releases backward compatible (in which case, shouldn't we say that somewhere)?

mikepizzo:
Fixes https://github.com/oasis-tcs/odata-specs/issues/2231
OData Clients should assume semantic versioning. OData services should be explicit.


mikepizzo:
Opened against wrong repo... :-(

